PR #79, opened 2019-10-15
Mg/test run func

mgoodhart5:
- Adds test for Run func
- Refactors Run func to pass in a `*kube.VPAClientInstance` instead of creating one inside the function
- Additional commented out test that will be in use after refactoring funcs in summary.go

### DIFF
--- a/pkg/summary/summary_test.go
+++ b/pkg/summary/summary_test.go
@@ -1,0 +1,94 @@
+// Copyright 2019 FairwindsOps Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


sudermanjr:
Added @rbren since he is usually gives good feedback on golang stuf. :-D

pkg/summary/summary_test.go
```diff
@@ -0,0 +1,106 @@
...
+	kubeClientVPA := kube.GetMockVPAClient()
+
+	fakeVpaListOptions := metav1.ListOptions{
+		LabelSelector: labels.Set(fakeVpaLabels).String(),
+	}
+
+	_, errOk := kubeClientVPA.Client.AutoscalingV1beta2().VerticalPodAutoscalers("").List(fakeVpaListOptions)
+	assert.NoError(t, errOk)
+
+	var summary Summary
```
rbren: Looks like we're just testing for an empty summary here?

Curious, why the call to `List`? Should we check that it returns something with `len() == 0`?

sudermanjr: I don't think that needs to be there. Not entirely sure. @mgoodhart5

mgoodhart5: Great question. We fixed this! Thanks!

PR #80, opened 2019-10-15
Add screenshot to readme

rbren:


README.md
```diff
@@ -11,6 +11,10 @@
 
 By using the kubernetes [vertical-pod-autoscaler](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler) in recommendation mode, we can see a suggestion for resource requests on each of our apps. This tool creates a VPA for each deployment in a namespace and then queries them for information.
 
+<div align="center">
+<img src="img/screenshot.png" alt="Goldilocks Screenshot" />
+<br>
+
```
mgoodhart5: I just viewed this file and it looks like this addition made everything below it also be center aligned. Does the <div> need to be closed?

rbren: Whoops good catch! Thanks.

Here's what the end result looks like: https://github.com/FairwindsOps/goldilocks/tree/rb/screenshot

mgoodhart5: Yay! It looks great!! ✨

README.md
```diff
@@ -11,6 +11,10 @@
 
 By using the kubernetes [vertical-pod-autoscaler](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler) in recommendation mode, we can see a suggestion for resource requests on each of our apps. This tool creates a VPA for each deployment in a namespace and then queries them for information.
 
+<div align="center">
```
sudermanjr: Can you put some text above this to explain what we are seeing here?  I don't think it's clear that this is the Goldilocks dashboard.  Perhaps it should be placed in a different spot?

rbren: Good point!

PR #87, opened 2019-10-23
Add flags to include/exclude namespaces

rbren:


rbren:
Still haven't tested this in a real-world setting - will do that once the images build.

e2e/test.sh
```diff
@@ -55,11 +59,15 @@ echo "********************************************************************"
 printf "\n\n"
 
 kubectl create ns goldilocks
+
 kubectl -n goldilocks apply -f /hack/manifests/dashboard/
 kubectl -n goldilocks apply -f /hack/manifests/controller/
 
-kubectl -n goldilocks wait deployment --timeout=60s --for condition=available -l app.kubernetes.io/name=goldilocks,app.kubernetes.io/component=dashboard
-kubectl -n goldilocks wait deployment --timeout=60s --for condition=available -l app.kubernetes.io/name=goldilocks,app.kubernetes.io/component=controller
+sleep $reinstall_wait
```
sudermanjr: I'm not sure why this is necessary since we have the kubectl wait below.  Can you shed some more light on that?

rbren: Whoops, not sure where that came from.

rbren:
Done!

Issue #82, opened 2019-10-17
Kind setup doesn't setup kubeconfig properly


sudermanjr:
The script hangs after cluster creation because KUBECONFIG doesn't seem to be getting set properly.

dosullivan:
Fixed in https://github.com/FairwindsOps/goldilocks/pull/89

Issue #81, opened 2019-10-17
Kind setup doesn't check version properly


sudermanjr:
https://github.com/FairwindsOps/goldilocks/blob/f318b5cb60de4ed6758a8cffecee08ff740b31e6/hack/kind/setup.sh#L25

This doesn't work correctly.

dosullivan:
Fixed in https://github.com/FairwindsOps/goldilocks/pull/89

PR #89, opened 2019-10-30
fix kind_version comparsion and kubeconfig in kind setup script

dosullivan:
In reference to https://github.com/FairwindsOps/goldilocks/issues/82 and https://github.com/FairwindsOps/goldilocks/issues/81

### DIFF
--- a/hack/kind/setup.sh
+++ b/hack/kind/setup.sh
@@ -74,4 +74,4 @@ if $install_goldilocks; then
   kubectl -n goldilocks apply -f ../manifests/dashboard
 fi
 
-echo "Use 'kind get kubeconfig-path --name=test-infra' to get your kubeconfig"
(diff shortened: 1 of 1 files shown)


hack/kind/setup.sh
```diff
@@ -37,7 +37,7 @@ kind create cluster \
   --image="kindest/$kind_node_image" || true
 
 # shellcheck disable=SC2034
-KUBECONFIG="$(kind get kubeconfig-path --name="test-infra")"
+export KUBECONFIG="$(kind get kubeconfig-path --name="test-infra")"
```
sudermanjr: I believe that unless this script is sourced, then the export won't get set outside the script. Do you think we should recommend sourcing it, or just output a recommendation on how to set the kubeconfig?

dosullivan: That's correct; I used `export` here to make the kubeconfig work within the script, so that it wouldn't get hung here. But I think you're right that we should recommend sourcing it the same way at the end of the script. I just updated the ending message to be more direct about that.

PR #99, opened 2019-12-11
Lots of small changes and cleanup

sudermanjr:
Updating kind.
Adding e2e for multiple kube versions.
Setting up pre-commit
Adding gcp auth and updating other modules


hack/kind/setup.sh
```diff
@@ -74,4 +73,4 @@ if $install_goldilocks; then
   kubectl -n goldilocks apply -f ../manifests/dashboard
 fi
 
-echo "Use 'export KUBECONFIG="$(kind get kubeconfig-path --name="test-infra")"' to set your kubeconfig to work with the kind cluster"
\ No newline at end of file
+echo "Your cluster should be running with goldilocks installed."
```
mgoodhart5: This wording makes me think that there is a possibility that it isn't running with goldilocks installed.

sudermanjr: I think that is entirely possible. :-D Especially if the install failed. I'll re-word to be more clear though.

Issue #78, opened 2019-10-15
Make units consistent


rbren:
If the units are currently set in e.g. `Mi`, we should round the recommendation (which seems to always be in `k` for memory).

This way, the dashboard will be more readable (no more `128Mi < 262144k`), and any copy/pasted config will stick to the user's current unit conventions.

<img width="357" alt="Screen Shot 2019-10-15 at 10 58 00 AM" src="https://user-images.githubusercontent.com/7611973/66843638-0fe0a400-ef3b-11e9-8def-418730216bb9.png">

sudermanjr:
This would be super annoying to implement.  This is the info that the Kube objects (vpa and deployment) return.  It's do-able, but the types for this are difficult to work with.

robscott:
It was definitely a pain to deal with this interface for kube-capacity. I'm not sure how easy it is to actually know the format the request was set in, but it's not that bad to just print out memory with Mi units like elsewhere in the k8s codebase. Here's the code behind [kubectl top](https://github.com/kubernetes/kubectl/blob/master/pkg/metricsutil/metrics_printer.go#L319) that formats resources, I used this same approach for kube-capacity for whatever that's worth.

rbren:
This should help as well: https://godoc.org/k8s.io/apimachinery/pkg/api/resource#Quantity.RoundUp

PR #103, opened 2020-01-16
Make units consistent for memory fixes issue #78

mgoodhart5:
Fixes issue #78 
Memory is now returned as Mi for recommendations
All tests passing

### DIFF
--- a/pkg/dashboard/helpers.go
+++ b/pkg/dashboard/helpers.go
@@ -23,7 +23,6 @@ import (
 )
 
 func printResource(quant resource.Quantity, resourceType corev1.ResourceName) string {
-	// weird thing you can't look at
 	if quant.IsZero() {
 		return "Not Set"
 	}
--- a/pkg/dashboard/helpers.go
+++ b/pkg/dashboard/helpers.go
@@ -15,16 +15,26 @@
 package dashboard
 
 import (
+	"fmt"
+
 	uuid "github.com/satori/go.uuid"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 )
 
-func printResource(quant resource.Quantity) string {
+func printResource(quant resource.Quantity, resourceType corev1.ResourceName) string {
+	// weird thing you can't look at
 	if quant.IsZero() {
 		return "Not Set"
 	}
-	return quant.String()
+	switch resourceType {
+	case corev1.ResourceCPU:
+		return fmt.Sprintf("%vm", quant.MilliValue())
+	case corev1.ResourceMemory:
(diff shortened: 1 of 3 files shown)


pkg/dashboard/helpers.go
```diff
@@ -15,16 +15,26 @@
...
+	"fmt"
+
 	uuid "github.com/satori/go.uuid"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 )
 
-func printResource(quant resource.Quantity) string {
+func printResource(quant resource.Quantity, resourceType corev1.ResourceName) string {
+	// weird thing you can't look at
```
sudermanjr: I don't think we need this comment ;-)

pkg/dashboard/helpers_test.go
```diff
@@ -29,31 +30,43 @@ func Test_getUUID(t *testing.T) {
 }
 
 func Test_printResource(t *testing.T) {
+	// struct to contain test cases defined below
```
sudermanjr: Unnecessary comments

pkg/dashboard/helpers.go
```diff
@@ -15,16 +15,26 @@
...
 	if quant.IsZero() {
 		return "Not Set"
 	}
-	return quant.String()
+	switch resourceType {
```
rbren: Minor nit: I think this switch duplicates `quant.String()`. This might be simpler:
```go
if resourceType == corev1.ResourceMemory {
  quant = quant.RoundUp(resource.Mega)
}
return quant.String()
```

pkg/dashboard/helpers_test.go
rbren: It took me a long time after graduating to accept this, but [comments are considered harmful](https://blog.humphd.org/vocamus-894/)  😅

pkg/dashboard/helpers.go
mgoodhart5: I am not sure I understand this, but happy to talk about it.

sudermanjr: Interesting. We were following the pattern from another example. Will that guarantee that `Mi` is returned rather than `k` for the units?  I am not familiar with the quant.RoundUp function.

sudermanjr: Also, what happens to CPU or other types of values in that case?

mgoodhart5: Not entirely sure why this doesn't want to work, but this is the error I got when I tried it.
`cannot use quant.RoundUp(resource.Mega) (type bool) as type resource.Quantity in assignment go`
Something with a type difference?

sudermanjr: Decided to just leave this for now. We can optimize later if we want

PR #120, opened 2020-03-03
Fix container current memory request

raspberrypython:


sudermanjr:
@raspberrypython Thanks for the contribution! Could you please rebase this on the latest master so we can get tests passing?

sudermanjr:
Looks like I was able to update the branch once I updated the branch protection settings

PR #137, opened 2020-03-16
Hkatz/add vpa update mode

hjkatz:
As I'm creating these PRs and pulling the commits over one at a time into new PRs on this new fork, I'm starting to wonder if it might be better to hand craft these PRs or re-implement some of them by hand and submit net-new changes as PRs rather than try to copy/paste commit by commit from our internal fork to this public fork.

I don't mind doing the copy/paste method, but I'm worried that sometimes the PRs may include accidental changes or not quite as clean code as the final product due to future commits that haven't been pulled in yet. For example in this PR you can see that in the last PR (#128) we had forgotten to enable the dry run feature in the controller, but in a commit in this PR we added it back in.

The net result is the same, but the process we take to get there is different.

What are y'all's thoughts on this? Personally the copy/paste method is very straightforward and pretty easy to do for me, so I like that. I also want to be clear and transparent about the changes we're putting up and proposing to implement in your project.

----

Note: This comment gives context to this PR: https://github.com/FairwindsOps/goldilocks/pull/128#issuecomment-599605258

----

* Add a VPAUpdateMode to the vpa.Reconciler

    The VPAUpdateMode field will allow us to use Goldilocks to create VPAs
    that operate outside of "off" mode.

* Add ability to specify vpa update mode

    This commit adds a parameter to createVPA that allows the user to
    specify the update mode for the VPA. The mechanism for specifying this
    is the label goldilocks.fairwinds.com/vpaUpdateMode.


sudermanjr:
> What are y'all's thoughts on this? Personally the copy/paste method is very straightforward and pretty easy to do for me, so I like that. I also want to be clear and transparent about the changes we're putting up and proposing to implement in your project.

I'm fine with the copy paste method. We aren't going to cut a release until all of these are in most likely, so it will be thoroughly tested via e2e and manually by me before we release. Whatever is easiest for you is fine with me, so copy paste is good.

pkg/vpa/vpa.go
```diff
@@ -118,7 +125,7 @@ func (r Reconciler) checkDeploymentLabels(deployment *appsv1.Deployment) (bool,
 	if len(deployment.ObjectMeta.Labels) > 0 {
 		for k, v := range deployment.ObjectMeta.Labels {
 			klog.V(7).Infof("Deployment Label - %s: %s", k, v)
-			if strings.ToLower(k) == "goldilocks.fairwinds.com/enabled" {
+			if strings.ToLower(k) == vpaEnabledLabel {
```
sudermanjr: We should consider switching these to use [ParseBool](https://golang.org/pkg/strconv/#ParseBool) instead so that more options will work.

pkg/vpa/vpa.go
```diff
@@ -258,3 +264,20 @@ func (r Reconciler) createVPA(namespace string, vpaName string) error {
 	}
 	return nil
 }
+
+func vpaUpdateModeForNamespace(ns *corev1.Namespace) v1beta2.UpdateMode {
```
sudermanjr: Am I reading this correctly that the updateMode is only configurable for a namespace? That's fine with me and expected based on the current configuration. I just wanted to check.

pkg/vpa/vpa.go
```diff
@@ -258,3 +264,20 @@ func (r Reconciler) createVPA(namespace string, vpaName string) error {
...
+	for k, v := range ns.GetLabels() {
+		if strings.ToLower(k) != vpaUpdateModeLabel {
+			continue
+		}
+		switch v {
```
sudermanjr: I noticed you have only implemented two of the four updateModes available. Is there a reason for that?

pkg/vpa/vpa.go
hjkatz: Oh yeah, that's a great idea! I'll go ahead and add this in a follow-up commit.

- [ ] Use `strcon.ParseBool` to verify boolean settings

pkg/vpa/vpa.go
hjkatz: Yes, like the other Goldilocks' managed labels on the Namespace we're sticking with the label == affects namespace pattern.

pkg/vpa/vpa.go
hjkatz: It looks like we only implemented exactly what we were looking to have access to. I also don't see us having added these additional update modes in the current master of our fork. I'll submit a commit to add the other modes too.

- [] Support all VPA update modes

pkg/vpa/vpa.go
hjkatz: I also updated `namespaceIsManaged` to use `ParseBool` too

hjkatz:
I've updated the PR with 3 additional commits based on your feedback. Let me know what you think! 👍

pkg/vpa/vpa.go
```diff
@@ -258,3 +259,28 @@ func (r Reconciler) createVPA(namespace string, vpaName string) error {
...
 }
+
+func vpaUpdateModeForNamespace(ns *corev1.Namespace) v1beta2.UpdateMode {
+	// See: https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1beta2/types.go#L101
+	updateMode := v1beta2.UpdateModeOff
+	for k, v := range ns.GetLabels() {
+		if strings.ToLower(k) != vpaUpdateModeLabel {
+			continue
+		}
+		switch v {
```
rbren: Should we just switch on `strings.ToLower(v)` instead of checking for different capitalizations?

hjkatz:
> Might be good to add a test case or two, but since this is an undocumented feature (and the default is tested) not a big deal.

I noticed that the test cases we've added for the reconciler aren't part of these commits, but do trust that we have added them and they'll be coming in a future PR. We have tests for just about all of the reconciler's different behaviours :)

pkg/vpa/vpa.go
hjkatz: Oh, that makes way more sense 🤦‍♂I'll update this

- [ ] Use `strings.ToLower(v)`

hjkatz:
Do y'all prefer that I click the "Resolve conversation" button when I submit additional commits to address comments? I've been using the checkmark/task thing in markdown

- [ ] like this :)

Either way, the additional commits have been added 👍

sudermanjr:
> Do y'all prefer that I click the "Resolve conversation" button when I submit additional commits to address comments? I've been using the checkmark/task thing in markdown
> 
> * [x]  like this :)
> 
> Either way, the additional commits have been added 👍

No preference. If they are resolved I just open them and review.

PR #140, opened 2020-03-24
Hkatz/vpa deployment exclusion and updates

hjkatz:
This PR the follow-up to #137 and #128 which adds the following:

- VPA Reconciler Tests (covering all new features that have been added)
- `vpa-opt-out` annotation for Deployments that would prefer to not have a VPA attached
  - This is particularly useful if the VPA update mode is "auto" for the Namespace.
  - The intent here is that goldilocks labels/configures whole Namespaces at a time, but individual Deployments may wish to be treated differently.
- Updates to logging and a bit of standardization for log levels and messages
  - This comes from hands on debugging and adding logs that actually helped while debugging behaviour.
- Update the reconciler to use real objects like VPA and Deployment and pass them around between functions instead of just names, this allows for more flexibility in the reconciler's functions.
- Add documentation to the README


hjkatz:
As an FYI, I honestly don't know why Github is saying there's 23 commits in this PR... there's only 5 that I can see:
```
* 563d78d (HEAD -> hkatz/vpa-deployment-exclusion-and-updates, origin/hkatz/vpa-deployment-exclusion-and-updates) Update README
* c57d004 Fix namespace logs to use Name instead of the full Namespace
* a7e18b2 Add more logging to VPA reconciler
* 662477b Change no-vpa annotation to vpa-opt-out (to opt out of vpa in auto mode)
* 65d1340 Add Deployment Exclusion Annotation
*   1e14372 (origin/master, origin/HEAD, master) Merge pull request #4 from FairwindsOps/master
```

README.md
```diff
@@ -141,6 +141,35 @@ You can set the default behavior for VPA creation using some flags. When specifi
...
+#### Enable Namespaces
+
+Namespaces are considered enabled or managed by goldilocks when the Namespace
+has the enabled label set to "true", for example:
+
+```
+kubectl label ns goldilocks goldilocks.fairwinds.com/enabled=true
+```
+
+#### VPA Update Mode
```
hjkatz: My reasoning here behind adding these docs is because the Namespace will already have the `vpa-update-mode` label visible and users may try fiddling with this without knowing how it works. At least with the docs we can explain what that label is, what it does, and that it shouldn't really be used.

pkg/vpa/vpa_test.go
```diff
@@ -332,3 +339,70 @@ func Test_ReconcileNamespaceRemoveLabel(t *testing.T) {
 	assert.Equal(t, 0, len(vpaList.Items))
 	assert.EqualValues(t, vpaList, &v1beta2.VerticalPodAutoscalerList{})
 }
+
+func Test_ReconcileNamespace_ExcludeDeploymentAnnotation(t *testing.T) {
```
hjkatz: I love the test harness that y'all set up btw. It makes it _very_ easy to add tests and ensure functionality!

README.md
```diff
@@ -141,6 +141,35 @@ You can set the default behavior for VPA creation using some flags. When specifi
...
+> Note: This feature is for advanced usage only and is not recommended nor the default!
+
+VPAs created for Deployments in a Namespace have an update mode of "off" by
+default, meaning the VPAs only report recommendations and do not actually
+auto-scale the Pods.
+
+The update mode can be changed for a namespace by labels as well, for example:
+
+```
+kubectl label ns goldilocks goldilocks.fairwinds.com/vpa-update-mode="aut"
```
sudermanjr:
```suggestion
kubectl label ns goldilocks goldilocks.fairwinds.com/vpa-update-mode="auto"
```

pkg/vpa/test_constants.go
```diff
@@ -74,3 +84,12 @@ var testDeployment = &appsv1.Deployment{
 		Name: "test-deploy",
 	},
 }
+
+var testDeploymentExcluded = &appsv1.Deployment{
+	ObjectMeta: metav1.ObjectMeta{
+		Name: "test-deploy",
+		Annotations: map[string]string{
```
sudermanjr: Curious why you chose to do annotations on the deployments when we use labels on the namespace? This seems like it might be confusing for the end-users.

pkg/vpa/vpa.go
```diff
@@ -162,69 +154,128 @@ func (r Reconciler) namespaceIsManaged(namespace *corev1.Namespace) bool {
...
+		if err != nil {
+			return err
+		}
+	} else {
+		// vpa exists
+		if *vpa.Spec.UpdatePolicy.UpdateMode == vpaUpdateMode {
+			klog.V(5).Infof("Deployment/%s does have a VPA/%s and matches vpa-update-mode=%s", deployment.Name, vpa.Name, vpaUpdateMode)
+		} else {
+			klog.V(5).Infof("Deployment/%s does have a VPA/%s but vpa-update-mode is different, updating from %s to %s", deployment.Name, vpa.Name, *vpa.Spec.UpdatePolicy.UpdateMode, vpaUpdateMode)
+			// vpa update mode does not matched what the namespace is configured for, update the VPA
```
sudermanjr:
```suggestion
			// vpa update mode does not match what the namespace is configured for, update the VPA
```

pkg/vpa/test_constants.go
```diff
@@ -74,3 +84,12 @@ var testDeployment = &appsv1.Deployment{
...
+			"goldilocks.fairwinds.com/vpa-opt-out": "true",
```
sudermanjr: Not necessarily for this PR, but with the update mode being configurable, I think maybe this should be a `vpa-update-mode-override` or something like that. Or maybe a pair of labels/annotations. One for update mode and one to just opt out entirely. 

Looking further below, I am a little confused as to the intent of this particular annotation. The README says `If you do not want a specific Deployment to have a VPA you can annotate the deployment`, but in the reconciler, you just set the update mode to `off`. I think we need a little more clarity around the intent and functionality of this/these annotations for the deployments

pkg/vpa/vpa.go
```diff
@@ -162,69 +154,128 @@ func (r Reconciler) namespaceIsManaged(namespace *corev1.Namespace) bool {
...
-	vpaDelete := utils.Difference(vpaNames, deploymentNames)
-	klog.V(5).Infof("Diff vpas, deployments: %v", vpaDelete)
+	return nil
+}
+
+func (r Reconciler) reconcileDeploymentAndVPA(nsName string, deployment appsv1.Deployment, vpa *v1beta2.VerticalPodAutoscaler, vpaUpdateMode v1beta2.UpdateMode) error {
+	// if the Deployment is opted out of vpa scaling then use vpa-update-mode="off"
+	if val, ok := deployment.GetAnnotations()[deploymentExcludedAnnotation]; ok {
+		if val == "true" {
+			klog.V(5).Infof("Deployment/%s has opted out of VPA scaling and will use vpa-update-mode=off", deployment.Name)
```
sudermanjr: This is the section that confused me about the intended functionality of the opt-out annotation. (see above comment)

pkg/vpa/test_constants.go
hjkatz: The whole labels vs. annotations thing is something I'm not quite sure of in terms of best practices from a kubernetes stand point too. From my understanding labels should be used sparingly to be used in a selector manner, and annotations are more free form in their purpose. I form this opinion mostly from here: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/

> You can use either labels or annotations to attach metadata to Kubernetes objects. Labels can be used to select objects and to find collections of objects that satisfy certain conditions. In contrast, annotations are not used to identify and select objects. The metadata in an annotation can be small or large, structured or unstructured, and can include characters not permitted by labels.

So in all honesty I'm not even sure we should have vpa mode or other things as labels. Overall it's your product and decision though :)

pkg/vpa/test_constants.go
hjkatz: Yes, this is a good point. I think the purpose behind this opt-out is to opt-out specifically from having the VPA in "auto" mode or having the VPA in an "updater" mode that would actually evict pods and apply recommendations. Perhaps I can make that clearer in the documentation?

pkg/vpa/test_constants.go
sudermanjr: I think you're probably right about annotations vs labels. However at this point we have gone all in on labels, and now is probably not the time to change this.  

there's probably a world where we switch to using annotations, but it will require supporting both for a time I think

pkg/vpa/test_constants.go
sudermanjr: Clearer documentation is definitely good. I think maybe a different name for it as well.

Perhaps a `goldilocks.fairwinds.com/vpa-mode: off` or something like that. This enables further customization down the road if we want to enable more of this behavior

hjkatz: I'll implement this later today

- [ ] Implement more specific control for Deployment annotation

hjkatz: "today" turned into Monday, whoops! Let me know what you think of the updated PR.

pkg/vpa/test_constants.go
sudermanjr: As a team we have decided that long-term annotations are the way to go. 

With that in mind, I opened #148 to track that. We will implement a graduated deprecation to do that.

For now, the deployments using annotations is fine.

hjkatz: Great! I didn't mean to move y'all's cheese, sorry about that :)

sudermanjr: Not a problem. I think it actually came up before, but we punted.

pkg/vpa/test_constants.go
```diff
@@ -74,3 +84,12 @@ var testDeployment = &appsv1.Deployment{
 		Name: "test-deploy",
 	},
 }
+
+var testDeploymentExcluded = &appsv1.Deployment{
+	ObjectMeta: metav1.ObjectMeta{
+		Name: "test-deploy",
+		Annotations: map[string]string{
+			"goldilocks.fairwinds.com/vpa-mode": "off",
```
rbren: Should this be `vpa-update-mode`?

pkg/vpa/vpa.go
```diff
@@ -162,69 +155,126 @@ func (r Reconciler) namespaceIsManaged(namespace *corev1.Namespace) bool {
...
+func (r Reconciler) reconcileDeploymentsAndVPAs(nsName string, vpas []v1beta2.VerticalPodAutoscaler, deployments []appsv1.Deployment, vpaUpdateMode v1beta2.UpdateMode) error {
+	// these keys will eventually contain the leftover vpas that do not have a matching deployment associated
+	vpaHasAssociatedDeployment := map[string]bool{}
+	for _, deployment := range deployments {
+		var dvpa *v1beta2.VerticalPodAutoscaler
+		// search for the matching vpa (will have the same name)
+		for _, vpa := range vpas {
+			if deployment.Name == vpa.Name {
+				// found the vpa associated with this deployment
+				dvpa = &vpa
```
rbren: This currently works, but is dangerous. It grabs a pointer to the temp variable `vpa`, rather than the array element. If you didn't `break`, it'd end up with the wrong value (the last element of `vpas`).

Instead, I'd suggest
```
dvpa = &vpas[idx]
```

pkg/vpa/vpa.go
```diff
@@ -162,69 +155,126 @@ func (r Reconciler) namespaceIsManaged(namespace *corev1.Namespace) bool {
...
+		}
+		klog.V(2).Infof("Reconciling Namespace/%s for Deployment/%s with VPA/%s in vpa-update-mode=%s", nsName, deployment.Name, vpaName, vpaUpdateMode)
+		err := r.reconcileDeploymentAndVPA(nsName, deployment, dvpa, vpaUpdateMode)
+		if err != nil {
+			return err
+		}
+	}
+
+	for _, vpa := range vpas {
+		if _, ok := vpaHasAssociatedDeployment[vpa.Name]; !ok {
```
rbren: I would suggest just using `if !vpaHasAssociatedDeployment[vpa.Name]`. It will default to `false` if `vpa.Name` isn't present.

That way if the logic above ever changes and we set `vpaHasAssociatedDeployment[vpa.Name] = false` somewhere, this block will still trigger

pkg/vpa/vpa.go
```diff
@@ -162,69 +155,126 @@ func (r Reconciler) namespaceIsManaged(namespace *corev1.Namespace) bool {
...
-	} else if len(vpaNeeded) > 0 {
-		for _, vpaName := range vpaNeeded {
-			err := r.createVPA(nsName, vpaName, vpaUpdateMode)
+		// for logging
+		vpaName := "none"
+		if dvpa != nil {
+			vpaName = dvpa.Name
+		}
+		klog.V(2).Infof("Reconciling Namespace/%s for Deployment/%s with VPA/%s in vpa-update-mode=%s", nsName, deployment.Name, vpaName, vpaUpdateMode)
+		err := r.reconcileDeploymentAndVPA(nsName, deployment, dvpa, vpaUpdateMode)
```
rbren: Rather than passing around `nsName`, I'd suggest just using `deployment.ObjectMeta.Namespace`.

pkg/vpa/vpa.go
```diff
@@ -162,69 +155,126 @@ func (r Reconciler) namespaceIsManaged(namespace *corev1.Namespace) bool {
...
+	if vpa == nil {
+		klog.V(5).Infof("Deployment/%s does not have a VPA currently, creating VPA/%s", deployment.Name, deployment.Name)
+		// no vpa exists, create one (use the same name as the deployment)
+		err := r.createVPA(nsName, deployment.Name, vpaUpdateMode)
+		if err != nil {
+			return err
+		}
+	} else {
+		// vpa exists
+		if *vpa.Spec.UpdatePolicy.UpdateMode == vpaUpdateMode {
```
rbren: Minor nit: can combine the two lines above into an `else if`. Would also suggest putting the `Update` block before the `No update` block.

pkg/vpa/vpa.go
```diff
@@ -162,69 +155,126 @@ func (r Reconciler) namespaceIsManaged(namespace *corev1.Namespace) bool {
...
+			klog.V(5).Infof("Deployment/%s does have a VPA/%s and matches vpa-update-mode=%s", deployment.Name, vpa.Name, vpaUpdateMode)
```
rbren: Maybe add something like `, leaving as-is` to the log, so it's obvious nothing is updating

pkg/vpa/vpa.go
```diff
@@ -162,69 +155,126 @@ func (r Reconciler) namespaceIsManaged(namespace *corev1.Namespace) bool {
...
 
-func (r Reconciler) listVPAs(namespace string) []string {
+func (r Reconciler) listDeployments(namespace string) ([]appsv1.Deployment, error) {
+	deployments, err := r.KubeClient.Client.AppsV1().Deployments(namespace).List(metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	klog.V(2).Infof("There are %d deployments in Namespace/%s", len(deployments.Items), namespace)
+	if klog.V(9) {
```
rbren: I'm surprised this compiles - shouldn't it be `if klog.V(9).Enabled`? Or can you magically use the return value as a boolean?

Ditto below

pkg/vpa/vpa.go
```diff
@@ -247,40 +297,69 @@ func (r Reconciler) createVPA(namespace, vpaName string, updateMode v1beta2.Upda
...
+	return nil
+}
+
+// vpaUpdateModeForResource searches the resource's annotations and labels for a vpa-update-mode
+// key/value and uses that key/value to return the proper UpdateMode type
+func vpaUpdateModeForResource(obj runtime.Object) v1beta2.UpdateMode {
+	var requestedVpaMode string
+
+	// check for vpa-update-mode in annotations first
+	accessor, _ := meta.Accessor(obj)
```
rbren: I did not know this was a thing! Very cool. This will help me in another project.

pkg/vpa/vpa.go
hjkatz: This realization was made further down the line and is coming in a future PR too.

pkg/vpa/vpa.go
rbren: Looks like you can indeed!

pkg/vpa/test_constants.go
hjkatz: Yes, fixed that.

- [ ] fix this

pkg/vpa/vpa.go
hjkatz: 👍 

- [ ] Fix this

pkg/vpa/vpa.go
hjkatz: 👍 [ ] Do this

pkg/vpa/vpa.go
hjkatz: This method is going to change significantly in a future PR where I switch to instead of updating the various fields of the vpa object directly to instead returning the expected vpa object in full. In order to keep the future PR less messy I'm going to leave this as is for now.

pkg/vpa/vpa.go
hjkatz: Same thing here, and I'll add a note to remember to update this wording in that PR.

hjkatz:
Sorry for the slow responses, it was a busy week last week. I'm able to focus more on the PRs for this week though :) I think up next is my dashboard overhaul :D

PR #154, opened 2020-04-13
Hkatz/update summary

hjkatz:
# Why?

The Summary package wasn't as flexible as needed for the next PR (there's a theme here, lol) for the changes that I'm planning to add for the Web UI and Dashboard packages. (Essentially splitting up the templates and providing different views (summaries) per Namespace, VPAs, etc...).

# What?

3 things are going on in this PR:

1. Reverted #90 (and ensured that the reusability goal behind this PR was still met)
1. Updated `pkg/summary` with new features, functionality, and tests
1. Standardized on the usage of `VPA` over `Vpa` as per golang best practices (this happened in this PR because in reality I had like 30 commits in one PR and I wanted to split them up and this one commit just happened to fit nicely here)


pkg/summary/options.go
```diff
@@ -0,0 +1,51 @@
+package summary
+
+import (
+	"github.com/fairwindsops/goldilocks/pkg/kube"
+	"github.com/fairwindsops/goldilocks/pkg/utils"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// Functional options
+type Option func(*Options)
```
rbren, line 10: Never seen this pattern before. Very cool

hjkatz: I found it a while back and have fallen in love with it. It's just so useful for all sorts of packages and `NewFoo()` functions 😄 Here's a blog post about it if you're interested: https://www.sohamkamani.com/golang/options-pattern/

pkg/summary/summary.go
```diff
@@ -49,119 +58,144 @@ type Summary struct {
...
+		if err != nil {
+			return summary, err
+		}
 	}
 
-	containerExclusions := strings.Split(excludeContainers, ",")
-
-	for _, vpa := range vpas.Items {
-		klog.V(8).Infof("Analyzing vpa: %v", vpa.ObjectMeta.Name)
+	if len(s.vpas) <= 0 {
```
lucasreed, line 101: Does this need to be `<=`? It feels like `== 0` would do the job.

pkg/summary/summary.go
```diff
@@ -49,119 +58,144 @@ type Summary struct {
...
 
-	summary.Namespaces = utils.UniqueString(summary.Namespaces)
+	// get the unique list of namespaces we've seen for this summary
+	summary.Namespaces = summaryNamespaces.List()
+
 	return summary, nil
 }
+
+// Update the list of VPAs that the summarizer uses
+func (s *Summarizer) UpdateVPAs() error {
```
lucasreed, line 169: Does this need to be a public method? It looks like the only caller is in this package.

pkg/summary/summary.go
```diff
@@ -49,119 +58,144 @@ type Summary struct {
...
+	return summarizer
 }
 
-func (client *Client) constructSummary(vpas *v1beta2.VerticalPodAutoscalerList, excludeContainers string) (Summary, error) {
+func (s Summarizer) GetSummary() (Summary, error) {
 	var summary Summary
-	if len(vpas.Items) <= 0 {
-		return summary, nil
+	// cached vpas
+	if s.vpas == nil {
```
lucasreed, line 94: It looks like the dashboard may be the only thing that makes use of the cache, but what happens if new VPAs get created after it creates the initial cache? Will the new ones get picked up? Should there be a TTL of the cache?

lucasreed: Actually looking closer at the dashboard code I don't think anything will ever cache because a new Summarizer object gets created on every call. See: https://github.com/FairwindsOps/goldilocks/pull/154/files#diff-dc1496f734707825dbb6070e68866784R167

pkg/dashboard/dashboard.go
```diff
@@ -162,7 +164,12 @@ func GetRouter(port int, basePath string, vpaLabels map[string]string, excludeCo
 			return
 		}
 
-		data, err := summary.GetInstance().Run(vpaLabels, excludeContainers)
+		summarizer := summary.NewSummarizer(
```
lucasreed, line 167: If we want to make use of the caching functionality in the `Summarizer` type, the `Summarizer` object should get passed into `GetRouter` from `cmd/dashboard.go`

pkg/summary/options.go
lucasreed: Agreed, I like this a lot!

pkg/summary/summary.go
```diff
@@ -20,13 +20,22 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/sets"
```
lucasreed, line 23: Cool, I haven't used this package before. I dig it.

pkg/summary/summary.go
```diff
@@ -49,119 +58,144 @@ type Summary struct {
...
+	// set the cached vpas list directly
+	summarizer.vpas = vpas
 
-	summary, _ := GetInstance().constructSummary(vpas, excludeContainers)
-	return summary, nil
+	return summarizer
 }
 
-func (client *Client) constructSummary(vpas *v1beta2.VerticalPodAutoscalerList, excludeContainers string) (Summary, error) {
+func (s Summarizer) GetSummary() (Summary, error) {
```
lucasreed, line 91: I think this should be a pointer
```suggestion
func (s Summarizer) GetSummary() (Summary, error) {
func (s *Summarizer) GetSummary() (Summary, error) {
```

pkg/summary/summary.go
```diff
@@ -49,119 +58,144 @@ type Summary struct {
...
+		return err
+	}
+	klog.V(10).Infof("Found vpas: %v", vpas)
+
+	s.vpas = vpas
+	return nil
+}
+
+// Run creates a summary of the vpa info for all namespaces.
+func (s Summarizer) listVPAs() ([]v1beta2.VerticalPodAutoscaler, error) {
```
lucasreed, line 187:
```suggestion
func (s *Summarizer) listVPAs() ([]v1beta2.VerticalPodAutoscaler, error) {
```

pkg/summary/summary.go
lucasreed: Or `< 1`

pkg/utils/utils.go
```diff
@@ -15,13 +15,13 @@
 package utils
 
 import (
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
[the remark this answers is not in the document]
hjkatz: Huh, looks like my auto-import thing in Vim added this for some reason. I'll remove it.

cmd/summary.go
```diff
@@ -36,12 +37,30 @@ func init() {
...
+	Use:   "summary [namespace]",
+	Short: "Generate a summary of the vpa recommendations in a namespace (default: all namespaces).",
+	Long: `Gather all the vpa data in a namespace and generate a summary of the recommendations.
+By default the summary will be about all VPAs in all namespaces.`,
+	Args: cobra.ArbitraryArgs,
 	Run: func(cmd *cobra.Command, args []string) {
+		opts := []summary.Option{}
+
+		// limit to a single namespace
+		if len(args) == 1 {
```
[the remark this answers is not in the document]
hjkatz: Sure! Added this in.

pkg/summary/summary.go
```diff
@@ -49,119 +58,144 @@ type Summary struct {
...
 
-// Run creates a summary of the vpa info for all namespaces.
-func (client *Client) Run(vpaLabels map[string]string, excludeContainers string) (Summary, error) {
-	klog.V(3).Infof("Looking for VPAs with labels: %v", vpaLabels)
-
-	vpaListOptions := metav1.ListOptions{
-		LabelSelector: labels.Set(vpaLabels).String(),
-	}
+// Returns a Summarizer for a known list of VPAs
+func NewSummarizerForVPAs(vpas []v1beta2.VerticalPodAutoscaler, setters ...Option) *Summarizer {
```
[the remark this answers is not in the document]
hjkatz: No I don't think so. This is a convenience function for dealing with in memory VPAs and getting information about them without going through the normal builder. I think I wrote it while intending to make use of it and I think the situation just never came up.

pkg/summary/summary.go
hjkatz: haha this is a habit of mine that I use to make sure that my if statements always cover _all_ cases and not just the cases I think could possibly happen. I can change this if y'all prefer a different style.

pkg/summary/summary.go
hjkatz: No, I catch this in an upcoming commit and change this to be private as well as revisit the caching behaviour of the Summarizer object to make a little more sense from a public API perspective.

pkg/summary/summary.go
hjkatz: This is what I think prompted me to change the caching and update behaviour of the Summarizer in the next set of commits where I go through and actually use the API in this package in the web ui and dashboard updates I make.

pkg/dashboard/dashboard.go
hjkatz: Absolutely! This is a shim for now so that the commits touched a more isolated package at one time. Next up is the dashboard updates (which also includes a revisit to the Summarizer package to assist the dashboard get the information it wants).

pkg/summary/summary.go
hjkatz: Why should this be a pointer receiver if the Summarizer struct shouldn't be modified during the function?

hjkatz:
@sudermanjr Looks like the e2e tests have a period difference in the kubectl responses coming back:

```
Expected: 'No resources found.'
Actual:   'No resources found'
```

Perhaps there's a way to use a regex for this test?

pkg/summary/summary.go
lucasreed: Fair point, I usually like to keep my API of a struct consistent across methods, but you are correct that it doesn't necessarily have to. It may just be an idiosyncrasy of mine that isn't necessary.

The only downside I could think of is if anything in the future was added in this function that would modify the struct, it may not be obvious when things don't work as planned.

pkg/summary/summary.go
lucasreed: I'll leave this and the other discussion above up to you, not a blocker.

pkg/summary/summary.go
lucasreed: Nah it's fine as is. It doesn't seem like it could ever be a negative number, but best not to make assumptions 😄

cmd/summary.go
```diff
@@ -36,12 +37,33 @@ func init() {
...
 
 var summaryCmd = &cobra.Command{
-	Use:   "summary",
-	Short: "Genarate a summary of the vpa recommendations in a namespace.",
-	Long:  `Gather all the vpa data in a namespace and generaate a summary of the recommendations.`,
+	Use:   "summary [namespace]",
+	Short: "Generate a summary of the vpa recommendations in a namespace (default: all namespaces).",
+	Long: `Gather all the vpa data in a namespace and generate a summary of the recommendations.
+By default the summary will be about all VPAs in all namespaces.`,
+	Args: cobra.ArbitraryArgs,
```
sudermanjr, line 44: Curious why you chose to use args instead of a namespace flag here, which would be consistent with other Kube-adjacent CLI tools.

sudermanjr, line 44: Curious why you chose to use args instead of a namespace flag here, which would be consistent with other Kube-adjacent CLI tools.

cmd/summary.go
```diff
@@ -36,12 +37,33 @@ func init() {
...
 	Run: func(cmd *cobra.Command, args []string) {
+		var opts []summary.Option
+
+		// limit to a single namespace
+		if len(args) == 1 {
+			opts = append(opts, summary.ForNamespace(args[0]))
+		} else if len(args) > 1 {
+			// undefined argument behaviour
```
sudermanjr, line 52: I don't think this comment is necessary. The klog.Fatal is information enough here.

cmd/summary.go
```diff
@@ -36,12 +37,33 @@ func init() {
...
+			klog.Fatal("Too many namespace arguments provided!")
+		}
+
+		// exclude containers from the summary
```
sudermanjr, line 56: Same as above, comment is unnecessary IMO

pkg/summary/options.go
```diff
@@ -0,0 +1,51 @@
+package summary
+
+import (
+	"github.com/fairwindsops/goldilocks/pkg/kube"
+	"github.com/fairwindsops/goldilocks/pkg/utils"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// Functional options
```
sudermanjr, line 9: linter: 

```
pkg/summary/options.go:9:1: comment on exported type Option should be of the form "Option ..." (with optional leading article)
```

There's a few of these. I suggest running `golint ./...` on this PR.

pkg/summary/options.go
```diff
@@ -0,0 +1,51 @@
...
+type Option func(*Options)
+
+// internal options for getting and caching the Summarizer's VPAs
+type Options struct {
```
sudermanjr, line 13: If these are internal options only, then should the type be un-exported?

pkg/summary/summary_test.go
```diff
@@ -74,8 +57,8 @@ func TestRunSummary(t *testing.T) {
 			Namespace: "testing",
 		},
 		Spec: v1beta2.VerticalPodAutoscalerSpec{
-			TargetRef: &autoscaling.CrossVersionObjectReference{
-				APIVersion: "apps/v1",
+			TargetRef: &autoscalingv1.CrossVersionObjectReference{
+				APIVersion: "extensions/v1beta1",
```
sudermanjr, line 61: Why did you change the apiVersion of the deployment in this test? This apiVersion will be completely removed in kube 1.16

sudermanjr:
See #159  for the updates to the e2e

pkg/summary/summary.go
lucasreed: I am actually leaning towards the approach you went with instead. Resolving this. Thanks!

pkg/summary/summary.go
lucasreed: I am actually leaning towards the approach you went with instead. Resolving this. Thanks!

cmd/summary.go
hjkatz: TBH I didn't even think of using a flag here. I was thinking more in terms of how I might want to use the summary command and thought `goldilocks summary <namespace>` made sense to me, but I now think that `goldilocks summary --namespace <namespace>` also makes just as much sense.

Which would you prefer?

pkg/summary/options.go
hjkatz: I've changed this to not be exported.

pkg/summary/options.go
hjkatz: I've fixed up the golint suggestions.

pkg/summary/summary_test.go
hjkatz: I'm not exactly sure. I've changed this to "apps/v1".

cmd/summary.go
sudermanjr: I generally lean towards a flag myself.

pkg/summary/options.go
```diff
@@ -6,21 +6,20 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
-// Functional options
-type Option func(*Options)
+type option func(*options)
```
lucasreed, line 9: This should not have been unexported. It's used in the `cmd` package. Once it's exported again you'll need to update the references in the the `summary` package as well.

hjkatz: Whoops! Nice catch, sorry for that!

cmd/summary.go
hjkatz: 👍

hjkatz:
So I think github is stuck or something. It looks like the PR thing says there's 1 change requested by @lucasreed, but I can't find the request or the review. Do y'all have any ideas on what changes are still wanted by Github?

> Merging can be performed automatically once the requested changes are addressed.

lucasreed:
> So I think github is stuck or something. It looks like the PR thing says there's 1 change requested by @lucasreed, but I can't find the request or the review. Do y'all have any ideas on what changes are still wanted by Github?
> 
> > Merging can be performed automatically once the requested changes are addressed.

I clicked requested changes when re-reviewing for the `Options` type being unexported. I just needed to review again.

PR #180, opened 2020-06-28
Hkatz/dashboard and summary overhaul

hjkatz:
# Why?

This is one of the largest PRs I'll be submitting. It's big because it does a complete overhaul of the dashboard (now "web") package and summary packages to support the changes.

Goals:
* Support Namespace list view
* Support individual Namespace view
* Support a more logical structure for a Summary
* Add testing to Summary

# What?

Major Changes:
* `pkg/dashboard` -> `pkg/web` to support more than just "the dashboard" view as there is also a list view too
* "dashboard" -> "web" in general
* Split templates by kind
* Templates and Handlers are more general purpose and can support future additions easily
* Adjust Summary structure to be more logically consistent with what it's summarizing (Summary -> Namespaces -> Deployments -> Containers)

Minor Changes:
* Allow web ClusterRole to list Namespaces
* Typo fixes, Squarespace fixes (stuff pulled in by accident)
* `/namespaces` is the new default route
* Test fixes, Typo Fixes, etc...
* Add outer structure to Summary
* Move commonly used labels to `pkg/utils`

P.S. After this there are a few more PRs but they're all 1-2 commits in size and should be quick. They include:
* Batch Summary Deployments list into a single API call to improve dashboard speed significantly
* Add Namespace list search bar/filtering and other friendly js improvements
* Add Namespace to blank Summary to support specific Namespaces in a more consistent structure
* Some final VPA touchups to make the business logic of updating VPAs clearer to modify and test
* Minor typos and QoL changes to catch some bugs (these will be batched into 1 PR)

This PR was already getting enormous so I'm trying to slice it here. I'm comfortable modifying just about anything in this PR now since the hard part was really getting past this point :)


pkg/kube/client.go
```diff
@@ -18,8 +18,9 @@ import (
 	"sync"
 
 	"k8s.io/client-go/kubernetes"
-	// Empty import needed for GCP auth. See client-go documentation
+	// Empty imports needed for supported auth methods in kubeconfig. See client-go documentation
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
```
hjkatz, line 23: I added this because locally we use oidc to auth to our clusters. I'm fine removing this change though and just putting it in locally when I want to test the UI and such.

pkg/web/assets/css/dashboard.css
```diff
@@ -12,6 +12,18 @@
   white-space: nowrap;
   overflow: hidden;
   margin-left: -2px; /* Fix for kerning issue */
+  display: inline;
```
hjkatz, line 15: I'm no UI or CSS professional so any adjustments here are welcomed :)

hjkatz:
I literally cannot believe all the e2e tests passed. That's incredible! :D

hjkatz:
@lucasreed/@sudermanjr Any chance y'all could lend some 👀  to this PR? 😃

sudermanjr:
> @lucasreed/@sudermanjr Any chance y'all could lend some 👀 to this PR? 😃

I will try to get a chance to look at this this week.

pkg/kube/client.go
rbren: Neat - I don't see any reason to remove. I wonder if we should add it to our other projects.

pkg/web/health.go
```diff
@@ -0,0 +1,22 @@
...
+		_, err := w.Write([]byte(healthyMessage))
+		if err != nil {
+			klog.Errorf("Error writing healthcheck: %v", err)
+		}
+	})
+}
+
+// Healthz replies with a zero byte 200 response
+func Healthz() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {})
```
rbren, line 21: does not writing anything to `w` just make it a 200? neat

pkg/kube/client.go
hjkatz: I actually found out recently that client-go can include all plugins with a single import. Let me know if y'all would like me to commit this instead.

```go
_ "k8s.io/client-go/plugin/pkg/client/auth"
```

Reference: https://github.com/kubernetes/client-go/blob/master/examples/out-of-cluster-client-configuration/main.go#L33-L40

cmd/web.go
```diff
@@ -0,0 +1,53 @@
...
+	rootCmd.AddCommand(webCmd)
+	webCmd.PersistentFlags().IntVarP(&serverPort, "port", "p", 8080, "The port to serve the web ui on.")
+	webCmd.PersistentFlags().StringVar(&basePath, "base-path", "/", "Path on which the web ui is served")
+	webCmd.PersistentFlags().StringVarP(&excludeContainers, "exclude-containers", "e", "", "Comma delimited list of containers to exclude from recommendations.")
+}
+
+var webCmd = &cobra.Command{
+	Use:   "web",
+	Short: "Run the goldilocks web ui that will show recommendations.",
+	Long:  `Run the goldilocks web ui that will show recommendations.`,
```
hjkatz, line 42:
```suggestion
	Short: "Run the goldilocks web UI that will show recommendations.",
	Long:  `Run the goldilocks web UI that will show recommendations.`,
```

CONTRIBUTING.md
```diff
@@ -49,7 +49,7 @@ Prerequisites:
 * Install the project with `go get github.com/fairwindsops/goldilocks`
 * Change into the goldilocks directory which is installed at `$GOPATH/src/github.com/fairwindsops/goldilocks`
 * Use `make tidy` or `make build` to ensure all dependencies are downloaded.
-* See the dashboard with `go run main.go dashboard`, then open http://localhost:8080/.  This assumes that you have a working KUBECONFIG in place with access to a cluster.
+* See the web ui with `go run main.go web`, then open http://localhost:8080/.  This assumes that you have a working KUBECONFIG in place with access to a cluster.
```
hjkatz, line 52:
```suggestion
* See the web UI with `go run main.go web`, then open http://localhost:8080/.  This assumes that you have a working KUBECONFIG in place with access to a cluster.
```

sudermanjr:
Can you explain why you decided it was necessary to change `dashboard` to `web`?  IMO this is a confusing change

pkg/kube/client.go
sudermanjr: If that works, might as well do all the auth. Although, it's really designed to be run in-cluster, so it's a little weird to support all this external auth just for testing locally

pkg/web/assets/css/dashboard.css
sudermanjr: I will probably get someone else to help with a UI pass at this before we release so that we can clean it up.  I'm not a huge fan of the overview page style choices.

pkg/kube/client.go
lucasreed: FWIW I've had issues in another project where importing the Azure auth broke my other k8s dependencies. If it works, go for it, just be aware it can potentially have unwanted side effects with deps.

lucasreed:
> I'm really not a fan of the dashboard -> web rename. It seems to suggest that it does more than display information, which it does not do.

>The fact that it also changes the name of all of the kubernetes objects means we will have to completely change the chart, as well as the users experiencing a very heavy shift in what they see in their clusters when deploying.

>TL;DR, it seems like a lot of code churn with zero functional purpose.

Agreed, simply changing the package name back to `dashboard` would make the PR much smaller which would be nice. A lot of secondary changes are only because of the name change (i.e. CONTRIBUTING + README)

hjkatz:
> > I'm really not a fan of the dashboard -> web rename. It seems to suggest that it does more than display information, which it does not do.
> 
> > The fact that it also changes the name of all of the kubernetes objects means we will have to completely change the chart, as well as the users experiencing a very heavy shift in what they see in their clusters when deploying.
> 
> > TL;DR, it seems like a lot of code churn with zero functional purpose.
> 
> Agreed, simply changing the package name back to `dashboard` would make the PR much smaller which would be nice. A lot of secondary changes are only because of the name change (i.e. CONTRIBUTING + README)

Will do y'all! I originally wanted to rename `pkg/dashboard` when I found out that it was going to hold more than _just_ a single dashboard page and it felt weird to have `dashboard.Dashboard()` as well as `dashboard.NamespaceList()`. I'll revert back to `pkg/dashboard` and we can move forward from there.

pkg/kube/client.go
hjkatz: We can leave this as is or I'm also fine taking this line out and only putting it in when I need to test locally. Up to y'all really.

hjkatz:
Ok, `pkg/web` has been un-renamed back to `pkg/dashboard`. The PR diff looks a lot smaller now. The file `pkg/dashboard/dashboard.go` diff is a bit hard to read because git is trying to be smart but it's not pattern matching well on the diff chunks, so I recommend viewing that file alone or in the side-by-side view. @sudermanjr @lucasreed

pkg/kube/client.go
sudermanjr: Let's just leave it as-is

PR #195, opened 2020-08-11
Hkatz/namespace and summary niceties

hjkatz:
# Why?

These are some minor improvements to the overall usage of the new namespace list, dashboard, and summary that I decided to include in a separate PR than the originally very large one.

# What?

* Add a search bar/filter with some nice js (like select on `<enter>` and such)
* Update the summarizer to produce valid summaries for single namespaces
* Only include containers in a summary if they match a container in the deployment (this fixes a bug where a summary would display blank containers, which were containers that don't match the deployment (this happens if a container is injected into the deployment))


pkg/dashboard/templates/namespace_list.gohtml
```diff
@@ -17,12 +18,22 @@
       <a class="button" href="/dashboard">View All</a>
     </div>
     <hr>
-    {{ range .Data }}
-    <div class="card namespace">
-      <h2>Namespace: <strong>{{.Name}}</strong></h2>
-      <a class="button" href="/dashboard/{{.Name}}">View</a>
+    <div class="search">
+      <input id="search"
```
sudermanjr, line 22: Can you please add a label element here for accessibility? Something like this:

```suggestion
      <label for="search">namespace search</label>
      <input id="search"
```

pkg/dashboard/assets/js/namespace_list.js
```diff
@@ -0,0 +1,37 @@
...
+    // filter namespace cards on every key typed
+    $searchBox.on("keyup", filterNamespaceCards)
+
+    // visit the first filtered namespace card's view/dashboard page by pressing enter
+    $searchBox.on("keypress", function(e) {
+        var key = e.which;
+        if(key == 13) { // enter
+            viewDashboardForFirstNamespace()
+        }
+    })
```
sudermanjr, line 33: I'm not really a huge fan of this functionality as it is now

There's no indication in the UI of this being how that works. When I type in a search field, I hit enter by default to do the search. This just jumps me somewhere with no warning or indication.

hjkatz:
> Can you please rebase this? I also left some small comments inline

This code is up to date. The reason we keep getting these seemingly extra commits is because we're merging PRs directly from `hjkatz/hkatz/feature` -> `FairwindsOps/master` and it's confusing github. I tried looking for a solution but didn't find anything offhand. Any ideas what to try?

pkg/dashboard/assets/js/namespace_list.js
hjkatz: I'm not sure I'm seeing the same behaviour that you're seeing. When I type in a Namespace I see the list filtering and then when I hit enter the top result is selected and the dashboard navigates to that Namespace's summary page. Alternatively when I use the mouse to click on a Namespace in the list the dashboard also navigates to that Namespace's summary page.

I'm not a UI designer though, so any suggestions are willingly accepted :)

sudermanjr: That's exactly the behavior I'm seeing, and it's very unexpected. When I hit enter, I expect to see search results, not be taken to the top result. 

There's really nothing to tell me what just happened or that it is going to happen when I hit enter.

sudermanjr:
The history on your branch is quite strange.  I would try `git rebase -i <Fairwinds Remote>/master` to get it back inline.

sudermanjr:
Also, it's not just extra commits on this one, it's extra changes that were included in the last PR

hjkatz:
> Also, it's not just extra commits on this one, it's extra changes that were included in the last PR

Agreed, I'm seeing the extra commits now. I'm going to close this PR and retry creating it tomorrow. I believe I applied an extra patch (from the previous PR) by accident.

pkg/dashboard/assets/js/namespace_list.js
hjkatz: We can drop the "visit first namespace on enter" behaviour entirely if you'd like. Then `<enter>` would effectively do nothing and would instead just rely on the filtering. It would be like an illusion of enter.

PR #196, opened 2020-08-12
Hkatz/namespace and summary niceties

hjkatz:
Re-creating PR#195

---

# Why?

These are some minor improvements to the overall usage of the new namespace list, dashboard, and summary that I decided to include in a separate PR than the originally very large one.

# What?

* Add a search bar/filter with some nice js (like select on `<enter>` and such)
* Update the summarizer to produce valid summaries for single namespaces
* Only include containers in a summary if they match a container in the deployment (this fixes a bug where a summary would display blank containers, which were containers that don't match the deployment (this happens if a container is injected into the deployment))


pkg/dashboard/assets/js/namespace_list.js
```diff
@@ -0,0 +1,37 @@
...
+    // filter namespace cards on every key typed
+    $searchBox.on("keyup", filterNamespaceCards)
+
+    // visit the first filtered namespace card's view/dashboard page by pressing enter
+    $searchBox.on("keypress", function(e) {
+        var key = e.which;
+        if(key == 13) { // enter
+            viewDashboardForFirstNamespace()
+        }
+    })
```
hjkatz, line 33: This behaviour is still up for discussion. See the original discussion here: https://github.com/FairwindsOps/goldilocks/pull/195#discussion_r468875157

@sudermanjr To quote the last part of the discussion and bring it over to this PR:

> We can drop the "visit first namespace on enter" behaviour entirely if you'd like. Then `<enter>` would effectively do nothing and would instead just rely on the filtering. It would be like an illusion of enter.
--Harrison

Thoughts?

sudermanjr: I think I would rather that than the way it is currently implemented. Other options would be to add a message that tells you what's happening, or maybe highlight the top result to indicate "selection".

To be 100% honest and up-front, this section is most likely going to get re-written in the near future anyway, so dropping it might be best.

hjkatz: Sure no worries! I'll remove this feature entirely as it's really not needed :)

sudermanjr: thanks!

PR #198, opened 2020-08-14
support all known auth methods

rbren:


sudermanjr:
I am doing a big PR to update dependencies. I'll just update this there.

PR #204, opened 2020-08-17
Fix use of labels for update mode

rbren:
This fixed a bug found by a user, who was trying to use a deployment's labels to set update mode.

### DIFF
--- a/pkg/vpa/vpa.go
+++ b/pkg/vpa/vpa.go
@@ -153,6 +153,7 @@ func (r Reconciler) namespaceIsManaged(namespace *corev1.Namespace) bool {
 }
 
 func (r Reconciler) reconcileDeploymentsAndVPAs(ns *corev1.Namespace, vpas []vpav1.VerticalPodAutoscaler, deployments []appsv1.Deployment) error {
+	defaultUpdateMode, _ := vpaUpdateModeForResource(ns)
 	// these keys will eventually contain the leftover vpas that do not have a matching deployment associated
 	vpaHasAssociatedDeployment := map[string]bool{}
 	for _, deployment := range deployments {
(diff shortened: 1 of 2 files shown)


sudermanjr:
You'll need to wait for me to merge #203 so that the e2e tests will work. Also can you please sign the CLA (you shouldn't need to since you're in the org, but :shrug: )

rbren:
Ahh thanks for the heads up, couldn't figure out what I'd broken

sudermanjr:
> Ahh thanks for the heads up, couldn't figure out what I'd broken

The demo chart we install as part of the e2e test needed updating because their image disappeared from docker hub :-(

sudermanjr:
@rbren would you mind rebasing this now that client-go is updated and the e2e tests are fixed?

rbren:
Done. Was kinda tricky so worth taking another look, but I think I got it.

pkg/vpa/vpa.go
```diff
@@ -338,7 +335,7 @@ func (r Reconciler) getVPAObject(existingVPA *vpav1.VerticalPodAutoscaler, ns *c
 			Name:       vpaName,
 		},
 		UpdatePolicy: &vpav1.PodUpdatePolicy{
-			UpdateMode: vpaUpdateModeForResource(ns),
+			UpdateMode: updateMode,
```
lucasreed, line 338:
```suggestion
			UpdateMode: vpaUpdateMode,
```

updateMode is undeclared. I'm only guessing at the name of the actual value based on what else is in this function. Please have a look and make sure.

lucasreed (changes requested):

pkg/vpa/vpa.go
rbren: Whoops, thanks, fixed

sudermanjr (approved):

lucasreed (approved):
This is just a rubber stamp, but approving with CI passing.

PR #209, opened 2020-08-24
Hkatz/vpa updates more tests and pointers

hjkatz:
# Why?

Here's the last PR from the set of changes Squarespace has made in their local fork. This last set of changes aims to improve the simplicity of the create/update/delete process for VPAs, Deployments, and Namespaces by using the real objects instead of string names throughout the reconciliation process. This comes with the added advantage of being able to test/mock these individual calls and methods.

# What?

* Use Namespace, VPA, and Deployment objects instead of string names
* Always update VPAs during reconciliation for simplicity of the reconciler and ensuring the VPAs stay in sync
* Use a helper func `getVPAObject()` to return an in-time state representation of the VPA and then use that for create/update/delete as needed
* Add tests


hjkatz:
Not sure how there are conflicts... everything is updated on my end. I'm going to fiddle with git and try to figure out what's going on and may close/reopen this PR.

hjkatz:
> Not sure how there are conflicts... everything is updated on my end. I'm going to fiddle with git and try to figure out what's going on and may close/reopen this PR.

Ok I figured out that I was missing the v1beta2 -> v1 changes so I've merged those.

PR #237, opened 2020-10-14
Fix typo shoul -> should

bharathappali:
Expecting `shoul` as `should` in line 99. Sorry if its not.
Not raising an issue as its a very trivial change.

Signed-off-by: bharathappali bharath.appali@gmail.com


sudermanjr (approved):
Thanks

Issue #240, opened 2020-10-23
multiple clusters functionality for the dashboard


claas-fridtjof-lisowski:
We have multiple clusters and having the option to only have one dashboard were one can select the desired cluster would be helpful.
Otherwise I need x dashboards for each cluster a separate one.

Having the option to give a kube/.config file or a cluster registry endpoint, so that all clusters are accessed which are found, would be nice.

Are there any workarounds, how are other people solving this?

rbren:
Thanks for reaching out! This is a fairly typical use case, but can be really difficult to implement in practice.

We've actually built a platform, [Fairwinds Insights](https://www.fairwinds.com/insights), which allows orgs to aggregate results from Goldilocks (as well as other k8s auditing tools like Trivy and Polaris) across all their clusters. It's a paid product rather than OSS (since we host a backend to do the aggregation), but you can start a free trial via the link above.

claas-fridtjof-lisowski:
I will update this PR with the latest changes from the new release soon.

PR #255, opened 2020-12-08
Feature: Add ability to query multiple clusters present in the kubeconfig

claas-fridtjof-lisowski:
This PR adds the ability the show all clusters which are added to the kube config file, referenced by the --kubeconfig flag.  
If no --kubeconfig is present it falls back to the default InCluster Kubernetes client config.

There are probably different ways and solutions to get Kubernetes clusters and authentication information.  
We do it this way that we have a custom entrypoint which gets all clusters (from an API) and their corresponding ServiceAccount tokens + CA certificates which are stored in HashiCorp Vault and add them to the .kube/conf.  
That file which is then referenced in the startup parameters for goldilocks.

Mentioned in #240. 

Edit:
I've adjusted my changes to the new ui which was release with v3.1.0  
A cluster list, similar to the namespace list is introduced and a new endpoint `/clusters` is added.  
The links in the navbar are hidden if in-cluster config is used.


sudermanjr:
@claas-fridtjof-lisowski Thanks for this PR. Just got back into the office today, so I'll try and take a look this week. It's going to take a bit to get this reviewed.

Also, I believe the dashboard changes are going to conflict with another in-flight UI PR, so we will have to evaluate the ordering on that.

claas-fridtjof-lisowski:
@sudermanjr thanks a lot.
We're using the multi-cluster ui already and for us it's helpful but there's no time pressure. The main purpose for this PR is that give back to the community and it'll be probably also helpful for others.

claas-fridtjof-lisowski:
Do the codecoverage failures need to be fixed or is it only the license which has to be signed?

sudermanjr:
That diff is probably fine, but the CLA is definitely required.

sudermanjr:
Thanks for updating this! I'll try and get a review in this week.

pkg/dashboard/templates/namespace_list.gohtml
```diff
@@ -8,41 +8,67 @@
   <header class="layoutSidebar__sidebar">
     {{ template "navbar" . }}
 
-    {{ if gt (len .Data) 1 }}
+    {{ if gt (len .Data.Name) 1 }}
```
sudermanjr, line 11: I believe this was intended to prevent the filter from showing up if there was only one namespace, not if the name was of len>1.  Not sure why this was changed.

```suggestion
    {{ if gt (len .Data) 1 }}
```

PR #266, opened 2020-12-22
update dependencies

sudermanjr:


lucasreed:
Should we go to the newest version of the k8s modules? I think it's up to `0.20.1`

sudermanjr:
> Should we go to the newest version of the k8s modules? I think they are up to `0.20.1`

Yep, good call. Done

lucasreed (approved):

PR #269, opened 2021-01-11
Fix flag parsing and passing of excluded and included namespaces

### LINKED ISSUE #268: --exclude-namespaces with a list of ns does not work
This works(indeed excluded):
- --on-by-default
- --exclude-namespaces=openshift-apiserver

This does not (meaning both ns get vpa created while both should be excluded)
- --on-by-default
- --exclude-namespaces=openshift-apiserver,openshift-apiserver-operator

sudermanjr:
Fixes #268

Also fixes a bad link I found in the readme.

### DIFF
--- a/cmd/controller.go
+++ b/cmd/controller.go
@@ -33,10 +33,10 @@ var dryRun bool
 
 func init() {
 	rootCmd.AddCommand(controllerCmd)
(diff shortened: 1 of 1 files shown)


sudermanjr:
I will rebase this tomorrow to incorporate the other link fix

lucasreed (approved):
approved, just need conflict resolution since the other PR was merged.

PR #270, opened 2021-01-12
fix broken link in README.md

realnumber:
At line 22, the link is missing `https://`, which redirects to an incorrect url (`https://github.com/FairwindsOps/goldilocks/blob/master/goldilocks.docs.fairwinds.com`).


sudermanjr:
Thank you! I actually already noticed this and added it into #269

sudermanjr (approved):

Issue #236, opened 2020-10-08
Namespace List Incomplete


isugimpy:
After turning on the functionality to create VPAs in all namespaces, I noticed most of mine don't show up in the list.  Digging into the code, I discovered the namespace list only returns namespaces which contain the label.  This seems undesirable, as going into all namespaces returns everything.  Should this, instead, check if the label is false, and if it's true or unset enumerate the VPAs to decide if it should be returned?

sudermanjr:
Thanks for bringing this to our attention. The flag should probably just be added to the dashboard as well in order to disable that filtering.

isugimpy:
I've opened a pull request for this issue, hopefully it can be brought in and released fairly quickly, as it's a pretty small change for a big value add.

PR #308, opened 2021-04-05
#236: Add on-by-default option to dashboard

isugimpy:
This PR aims to fix the issue raised in #236 where namespaces must explicitly be enabled to show up.  Noteworthy, I'm not aware of a way to do a LabelSelector with a `!=` value using the map syntax that was already being used in namespace-list.go, so I constructed the selector manually.  If there's a more preferred way to handle this, I'm certainly open to updating it.  This should be ready to review.  I've tested it on a real cluster and it worked as expected.


lucasreed (approved):
@isugimpy Hey thanks a ton, this change LGTM, appreciate you taking the time 🎉 

Once tests past after bringing this branch up to date we'll get it merged.

PR #360, opened 2021-09-20
Update golangci-lint to v1.41.1

lucasreed:
There are a lot of bugfixes since the version we are currently using and I think the error we see on [this build](https://app.circleci.com/pipelines/github/FairwindsOps/goldilocks/957/workflows/0d837a86-bf8c-432a-8360-e6ad13a85480/jobs/3737) is due to a bug.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,7 +32,7 @@ jobs:
       - run:
           name: golangci-lint
           command: |
-            curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.23.6
+            curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.41.1
             golangci-lint run -v
       - run:
           name: Run Tests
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       - run:
           name: golangci-lint
           command: |
-            curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.23.6
+            curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.41.1
```
sudermanjr, line 35: Maybe we should move to a different way of downloading this. I'm not a huge fan of the unverified curl bash

lucasreed: It is unfortunately the recommended way to do it: https://golangci-lint.run/usage/install/#other-ci

sudermanjr: Can we pull it from releases and verify the sha?
https://github.com/golangci/golangci-lint/releases

lucasreed: Ok my recent change does that instead. I made the version and checksum env vars so it's not too cumbersome to update.

sudermanjr (approved):

PR #375, opened 2021-10-21
Vpa Resource Policy

kengibous:
If the annotation `goldilocks.fairwinds.com/vpa-resource-policy` exists on a namespace then `ResourcePolicy` section of the `VerticalPodAutoscalerSpec` will be defined.  The value for this annotation is expected to be the `json` data that defines the `PodResourcePolicy`


kengibous:
@lucasreed yes I will work on e2e test now

PR #385, opened 2021-11-11
Fix VPA installation command in docs

marcomicera:
The `gcloud container` command is not longer under `gcloud beta` anymore, and a mandatory `--region`/`--zone` flag was missing too.


lucasreed (approved):
Thank you!! 🎉

PR #393, opened 2021-11-24
[UPGRADE WARNING] Forces re-create of all VPA objects - Prefix vpa names with goldilocks- 

### LINKED ISSUE #205: Prefix VPAs
VPAs should be prefixed with a configurable prefix (e.g. `goldilocks-`) to avoid collisions with existing VPAs.

sudermanjr:
fixes #205

On upgrade to this version, all VPAs will be re-created with the names prefixed.


lucasreed (changes requested):
LGTM, my only request is that we put a note in the docs somewhere about upgrading to `v4.1.0` will make existing VPAs recreate.

sudermanjr:
@lucasreed do you think the title of this PR ( which will end up in the release notes) is sufficient for that warning?

Do we want to do a major version bump? That seems excessive since the actual behavior isn't changing and it is backward-compatible, but it might also be a good idea. Idk.

lucasreed (approved):

PR #423, opened 2022-02-05
hack/manifests: update rbac.authorization.k8s.io to v1

mlazowik:
Required for k8s v1.22+, docs say there are no notable changes to the api. See https://kubernetes.io/docs/reference/using-api/deprecation-guide/#rbac-resources-v122


lucasreed (approved):
Apologies for the delay here. I'm cool with this as they were deprecated back in 1.17 which is out of support. Thank you!

PR #492, opened 2022-07-14
update GKE notes section with difference between GKE Autopilot and St…

kimschles:
## Checklist
* [x] I have signed the CLA

To update the GKE notes in the installation instructions. 

I added a note about the difference between VPA in GKE autopilot and standard. VPA is no longer in Beta in GKE.

Please double check my work. I see the sentence "Vertical Pod autoscaling is enabled by default in Autopilot clusters." in the [GKE VPA docs](https://cloud.google.com/kubernetes-engine/docs/concepts/verticalpodautoscaler), but it's possible I'm not reading it right. 

Also, HI ANDY!!

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -25,7 +25,7 @@ The full VPA install includes the updater and the admission webhook for VPA. Gol
 
 ### GKE Notes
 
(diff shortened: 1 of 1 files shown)


sudermanjr:
Thanks! 


HI KIM

PR #516, opened 2022-08-31
Calculate the Costs

hieptl:
This PR fixes #

## Checklist
* [x] I have signed the CLA

* Calculate the costs after the users click on the `Save` button on the `Cost Settings` component.

* Updated the dashboard.go to calculate the cost

* None


sudermanjr (changes requested):
Looks like the linter is failing:

```
pkg/dashboard/dashboard.go:32:7: `megabyte` is unused (deadcode)
const megabyte = kilobyte * 1000
```

pkg/dashboard/dashboard.go
```diff
@@ -52,6 +66,47 @@ func Dashboard(opts Options) http.Handler {
...
+
+						c.TopNumber = topNumber
+						c.LowerNumber1 = math.Abs(lowerNumber1)
+						c.LowerNumber2 = math.Abs(lowerNumber2)
+
+						w.Containers[k] = c
+					}
+				}
+			}
+		}
```
sudermanjr, line 108: It would be really nice if this was split into multiple functions in order to reduce the overall complexity of this section. As it is now, it's very confusing, and I am struggling to understand what it's supposed to be doing. Additionally, if we split them out, then we can test these functions, which will add even more clarity.

pkg/dashboard/dashboard.go
```diff
@@ -15,19 +15,33 @@
...
 	"k8s.io/klog/v2"
 
 	"github.com/fairwindsops/goldilocks/pkg/summary"
 )
 
+const kibabyte = 1024
+const mebibyte = kibabyte * 1024
+const gibibyte = mebibyte * 1024
+const kilobyte = 1000
+const megabyte = kilobyte * 1000
```
sudermanjr, line 32: Nit: It's a bit cleaner to use this syntax (also include the other const down below. 

```suggestion
const (
    kibabyte = 1024
    mebibyte = kibabyte * 1024
    gibibyte = mebibyte * 1024
    kilobyte = 1000
    megabyte = kilobyte * 1000
)
```

sudermanjr (changes requested):

pkg/dashboard/dashboard.go
hieptl: I updated the code. Thank you!

pkg/dashboard/dashboard.go
hieptl: I updated the code. Thank you!

pkg/dashboard/dashboard.go
```diff
@@ -78,3 +115,54 @@ func Dashboard(opts Options) http.Handler {
 		writeTemplate(tmpl, opts, &data, w)
 	})
 }
+
+func calculateContainerCost(costPerCPUFloat float64, costPerGBFloat float64, c summary.ContainerSummary) float64 {
+	cpuCost := costPerCPUFloat * (c.Requests.Cpu().AsApproximateFloat64() + c.Limits.Cpu().AsApproximateFloat64()) / 2
+	memCost := costPerGBFloat * (ConvertToGB(c.Requests.Memory().Value()) + ConvertToGB(c.Limits.Memory().Value())) / 2
+	return toFixed(cpuCost+memCost, 4)
```
rbren, line 122: If `Limits` are not set, we should only use `Requests`

pkg/dashboard/templates/container.gohtml
```diff
@@ -111,6 +116,11 @@
     <a href="#burstable-qos"
       >Burstable <abbr title="Quality of Service">QoS</abbr></a
     >
+    {{if lt $.BurstableCostInt 0}}
+      <span class="lower-number lower-number--negative">-${{$.BurstableCost}}/hour</span>
+    {{else if gt $.BurstableCostInt 0}}
```
rbren, line 121: These aren't showing up for me. Maybe this should just be an `else` instead of an `else if`?

hieptl: I updated the code. Thank you

pkg/dashboard/dashboard.go
hieptl: I updated the code. Thank you

pkg/dashboard/dashboard.go
```diff
@@ -15,19 +15,33 @@
...
 
 	"github.com/gorilla/mux"
 	"k8s.io/klog/v2"
 
 	"github.com/fairwindsops/goldilocks/pkg/summary"
 )
 
+const (
+	kibabyte = 1024
+	mebibyte = kibabyte * 1024
```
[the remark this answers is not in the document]
hieptl: I updated the code. Thank you

pkg/dashboard/dashboard.go
```diff
@@ -78,3 +115,59 @@ func Dashboard(opts Options) http.Handler {
 		writeTemplate(tmpl, opts, &data, w)
 	})
 }
+
+func calculateContainerCost(costPerCPUFloat float64, costPerGBFloat float64, c summary.ContainerSummary) float64 {
+	cpuCost := costPerCPUFloat * (c.Requests.Cpu().AsApproximateFloat64() + c.Limits.Cpu().AsApproximateFloat64()) / 2
```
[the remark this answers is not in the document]
hieptl: I updated the code. Thank you

rbren:
Now when I don't have any settings in place, I see +$0
Screen Shot 2022-09-14 at 9 14 29 AM

pkg/dashboard/dashboard.go
```diff
@@ -78,3 +115,60 @@ func Dashboard(opts Options) http.Handler {
 		writeTemplate(tmpl, opts, &data, w)
 	})
 }
+
+func calculateContainerCost(costPerCPUFloat float64, costPerGBFloat float64, c summary.ContainerSummary) float64 {
+	var memCost, cpuCost float64
+	if c.Limits != nil {
+		cpuCost = costPerCPUFloat * (float64(c.Requests.Cpu().Value() + c.Limits.Cpu().Value())) / 2
+		memCost = costPerGBFloat * (ConvertToGB(c.Requests.Memory().Value()) + ConvertToGB(c.Limits.Memory().Value())) / 2
```
rbren, line 123: This is a little more complicated--the user might set `Limits.CPU`, but not `Limits.Memory` (or vice versa). Requests can also be nil or 0.

I would write the code like this:
```go
cpuRequests := 0
memRequests := 0
cpuLimits := 0
memLimits := 0

if c.Limits != nil {
  cpuLimits = c.Limits.Cpu().Value()
  memLimits = c.Limits.Memory().Value()
}
if c.Requests != nil {
  cpuRequests = c.Requests.Cpu().Value()
  memRequests = c.Requests.Memory().Value()
}

func getNonZeroAverage(req, limit float64) float64 {
  avg := (req + limit) / 2.0
  if req == 0.0 {
    avg = limit
  }
  if limit == 0.0 {
    avg = request
  }
  return avg
}

cpuCost := costPerCPUFloat * getNonZeroAverage(cpuRequests, cpuLimits)
memCost := costPerGBFloat * getNonZeroAverage(memRequests, memLimits)
```

pkg/dashboard/dashboard.go
```diff
@@ -78,3 +115,60 @@ func Dashboard(opts Options) http.Handler {
...
+}
+
+func calculateRecommendedCosts(costPerCPUFloat float64, costPerGBFloat float64, containerCost float64, c summary.ContainerSummary) (float64, float64) {
+	guaranteedCpuCostRecommended := costPerCPUFloat * float64(c.Target.Cpu().Value())
+	guaranteedMemCosttRecommended := costPerGBFloat * ConvertToGB(c.Target.Memory().Value())
+
+	burstableCpuCostRecommended := costPerCPUFloat * (float64(c.LowerBound.Cpu().Value() + c.UpperBound.Cpu().Value())) / 2
+	burstableMemCosttRecommended := costPerGBFloat * (ConvertToGB(c.LowerBound.Memory().Value()) + ConvertToGB(c.UpperBound.Memory().Value())) / 2
+
+	return toFixed(containerCost-(guaranteedCpuCostRecommended+guaranteedMemCosttRecommended), 4), toFixed(containerCost-(burstableCpuCostRecommended+burstableMemCosttRecommended), 4)
```
rbren, line 138: We should flip this subtraction. I.e. if recommendedCost > containerCost, this should be a positive number.

hieptl: I updated the code. Thank you

pkg/dashboard/dashboard.go
hieptl: I updated the code. Thank you for the suggestion!

pkg/dashboard/dashboard.go
rbren: @hieptl did you push your code? this still looks the same to me.

pkg/dashboard/dashboard.go
```diff
@@ -78,3 +115,84 @@ func Dashboard(opts Options) http.Handler {
...
+
+func calculateRecommendedCosts(costPerCPUFloat float64, costPerGBFloat float64, containerCost float64, c summary.ContainerSummary) (float64, float64) {
+	guaranteedCpuCostRecommended := costPerCPUFloat * float64(c.Target.Cpu().Value())
+	guaranteedMemCosttRecommended := costPerGBFloat * ConvertToGB(c.Target.Memory().Value())
+
+	burstableCpuCostRecommended := costPerCPUFloat * (float64(c.LowerBound.Cpu().Value() + c.UpperBound.Cpu().Value())) / 2
+	burstableMemCosttRecommended := costPerGBFloat * (ConvertToGB(c.LowerBound.Memory().Value()) + ConvertToGB(c.UpperBound.Memory().Value())) / 2
+
+	guaranteedCostRecommended := math.Abs(containerCost - (guaranteedCpuCostRecommended + guaranteedMemCosttRecommended))
+	burstableCostRecommended := math.Abs(containerCost - (burstableCpuCostRecommended + burstableMemCosttRecommended))
```
rbren, line 160: This is not quite right - we don't want an absolute value.

If containerCost > recommendedCost, it should be negative (i.e. you'll save money by taking the recommendations)

hieptl: I updated the code. Thank you so much!

rbren:
@hieptl I'm consistently getting r5n.large as my instance type...

pkg/dashboard/dashboard.go
```diff
@@ -156,10 +156,17 @@ func calculateRecommendedCosts(costPerCPUFloat float64, costPerGBFloat float64,
...
 
-	return toFixed(guaranteedCostRecommended, 4), toFixed(burstableCostRecommended, 4)
+	return transformRecommendedCost(containerCost, guaranteedCostRecommended), transformRecommendedCost(containerCost, burstableCostRecommended)
+}
+
+func transformRecommendedCost(containerCost float64, recommendedCost float64) float64 {
+	if containerCost > recommendedCost {
+		return -1.0 * toFixed(containerCost-recommendedCost, 4)
+	}
+	return toFixed(recommendedCost-containerCost, 4)
```
rbren, line 169: This is still not correct: -1 * (B-A) is the same as (A-B) 😄 

It should just be `recommendedCost - containerCost` - no need to multiply.

pkg/dashboard/assets/js/cost_settings.js
```diff
@@ -25,6 +27,8 @@
   const costPerGBKey = "costPerGB";
   const otherOption = "Other";
   const emptyString = "";
+  const defaultCloudProvider = "AWS";
+  const defaultInstanceType = "663";
```
rbren, line 31: This might be the issue--IDs are numbers, but this is a string. I think this should be `663`, not `"663"`

rbren: Oh maybe not...I see you're using `option.value`. I wonder if it depends on the browser whether `option.value` gets converted to a string? I'm in firefox

rbren (approved):
@hieptl I had to change the cpu calculations to `MilliValue` - `Value()` rounded things like `10m` up to a full CPU!

PR #524, opened 2022-09-26
Add Cost Functionality to Goldilocks

rbren:
This PR adds cost functionality to goldilocks

## Checklist
* [x] I have signed the CLA

Add the ability to show workload costs and cost changes

Lots of frontend updates, as well as some backend changes


transient1 (approved):

pkg/dashboard/templates/cost_settings.gohtml
```diff
@@ -0,0 +1,51 @@
...
+          </div>
+        </div>
+      </div>
+      <div class="cost-settings-box__inputs-container">
+        <label class="cost-settings-box__cost-per-cpu">
+          <span id="cost-settings-box__cost-per-cpu-label-content">Cost per CPU-hour</span>
+          <input id="cost-settings-box__cost-per-cpu-input" type="number" step="0.01" required disabled />
+        </label>
+        <label class="cost-settings-box__cost-per-gb">
+          <span id="cost-settings-box__cost-per-gb-label-content">Cost per GB-hour</span>
```
rbren, line 43:
```suggestion
          <span id="cost-settings-box__cost-per-gb-label-content">Dollars per GB-hour</span>
```

sudermanjr:
Can you please fill out the PR description?


Also, has this been tested with a screen reader for accessibility? 
There's some JavaScript that's showing/hiding things that make me worry about focus management, as well as the some of the CSS that may cause issues with accessibility.

sudermanjr (changes requested):
Please remove the banner ad for the benchmark report

sudermanjr (changes requested):
Once the cost data is added, every time I click on a namespace, the screen flashes several times as the javascript hides/shows elements. Is it possible to smooth this out?

pkg/dashboard/templates/cost_settings.gohtml
```diff
@@ -0,0 +1,51 @@
...
+      </div>
+    </summary>
+
+    <form class="cost-settings-box__content">
+      <p>Choose an AWS or GCP instance type to get an accurate cost data, or set your cost per CPU and cost per GB of memory manually. Pick "Other" Cloud Provider to customize the numbers.</p>
+      <div class="cost-settings-box__options-container">
+        <div class="cost-settings-box__cloud-providers">
+          <label for="cloud-providers">Cloud Provider</label>
+          <div class="cost-settings-box__select-container">
+            <select name="cloud-providers" id="cost-settings-box__cloud-providers">
```
sudermanjr, line 23: This label doesn't actually relate to the selector. You can see this in the accessibility section of the dev-tools inspector

pkg/dashboard/templates/cost_settings.gohtml
```diff
@@ -0,0 +1,51 @@
...
+            </select>
+            <img src="/static/images/triangle.svg" alt="Select Icon" />
+          </div>
+        </div>
+        <div class="cost-settings-box__instance-types">
+          <label for="instance-types">Instance Type</label>
+          <div class="cost-settings-box__select-container">
+            <select name="instance-types" id="cost-settings-box__instance-types">
+            </select> 
+            <img src="/static/images/triangle.svg" alt="Select Icon" />
```
sudermanjr, line 33: Since this is purely a visual enhancement, and isn't needed for usability, should it be hidden from screen readers?

rbren:
@hieptl when you have some time, can you address Andy's feedback above?

rbren:
Re: accessibility, I will see if I can enlist Ivan to help us surface any issues.

pkg/dashboard/templates/dashboard.gohtml
```diff
@@ -156,5 +159,11 @@
       {{ template "footer" . }}
     </footer>
   </div>
+  <script>
+    window.INSIGHTS_HOST = "{{ .Data.InsightsHost }}"
+  </script>
+  <script defer src="static/js/email.js"></script>
+  <script defer src="static/js/api-token.js"></script>
+  <script defer src="static/js/cost_settings.js"></script>
```
sudermanjr, line 167: Putting these here is an outdated pattern that may affect performance negatively on "large sites" which may be triggered when we have a large amount of data.

> An old-fashioned solution to this problem used to be to put your script element right at the bottom of the body (e.g. just before the </body> tag), so that it would load after all the HTML has been parsed. The problem with this solution is that loading/parsing of the script is completely blocked until the HTML DOM has been loaded. On larger sites with lots of JavaScript, this can cause a major performance issue, slowing down your site. 

From: https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/What_is_JavaScript#script_loading_strategies

pkg/dashboard/templates/cost_settings.gohtml
sudermanjr: (this applies to the other box as well)

pkg/dashboard/dashboard.go
```diff
@@ -52,18 +66,133 @@ func Dashboard(opts Options) http.Handler {
...
 	})
 }
+
+func calculateContainerCost(costPerCPUFloat float64, costPerGBFloat float64, c summary.ContainerSummary) float64 {
+	var cpuRequests, memRequests, cpuLimits, memLimits float64
+
+	cpuRequests = 0
+	memRequests = 0
+	cpuLimits = 0
+	memLimits = 0
```
sudermanjr, line 125: I believe the zero-value of a float in go is zero, and it's not nillable, so these are unnecessary after declaring the vars.

pkg/dashboard/templates/cost_settings.gohtml
```diff
@@ -0,0 +1,51 @@
...
+          ><i aria-hidden="true" class="fas fa-fw fa-angle-up --showWhenOpen"></i>
+        </p>
+        <div>
+         <button id="cost-settings__disable-cost-settings">Disable Cost Settings</button>
+        </div>
+      </div>
+    </summary>
+
+    <form class="cost-settings-box__content">
+      <p>Choose an AWS or GCP instance type to get an accurate cost data, or set your cost per CPU and cost per GB of memory manually. Pick "Other" Cloud Provider to customize the numbers.</p>
```
sudermanjr, line 18: This text doesn't seem to be read by the screenreader in my testing. It could be that I'm using it wrong, but I don't seem to be able to find this description.

pkg/dashboard/templates/cost_settings.gohtml
```diff
@@ -0,0 +1,51 @@
...
+        <div class="cost-settings-box__cloud-providers">
+          <label for="cloud-providers">Cloud Provider</label>
+          <div class="cost-settings-box__select-container">
+            <select name="cloud-providers" id="cost-settings-box__cloud-providers">
+            </select>
+            <img src="/static/images/triangle.svg" alt="Select Icon" />
+          </div>
+        </div>
+        <div class="cost-settings-box__instance-types">
+          <label for="instance-types">Instance Type</label>
```
sudermanjr, line 29: Same with this label, it doesn't select the cost settings box, so the screen reader never reads it.

pkg/dashboard/templates/cost_settings.gohtml
```diff
@@ -0,0 +1,51 @@
...
+  <details open>
+    <summary class="--hideMarker">
+      <div class="cost-settings__title-container">
+        <p class="cost-settings-box__toggle-title">
+          Cost Settings
+          <i aria-hidden="true" class="fas fa-fw fa-angle-down --showWhenClosed"></i
+          ><i aria-hidden="true" class="fas fa-fw fa-angle-up --showWhenOpen"></i>
+        </p>
+        <div>
+         <button id="cost-settings__disable-cost-settings">Disable Cost Settings</button>
```
sudermanjr, line 12: This button living inside of the summary element isn't semantic HTML because it puts a button inside of a button. The "Disable Cost Settings" button needs to be outside of this summary element.

You can notice the negative effects of this both visually when you go to click the button and it looks like you're collapsing the form, as well as when you use a screenreader and it reads "Cost settings - disable cost settings" all at once.

hieptl:
Hi @sudermanjr : I updated the code based on your feedback. Thank you for your feedback!

sudermanjr (approved):
Thanks for all the updates!

PR #528, opened 2022-10-02
Make namespace filter links unique

### LINKED ISSUE #505: "View only this namespace" links should be unique
### What happened?

The `namespace.gohtml` template, used to show the contents and details of any given namespace, includes a link that says "View only this namespace." It's possible that this link shows up multiple times on a single page, but navigates out to an entirely different view.

This is especially problematic in the context of screen readers. One useful feature of these tools is the ability to surface a list of all links on a page. Someone who used this functionality now would get back a repetitive and unhelpful list.

See [Understanding Success Criterion 2.4.9: Link Purpose (Link Only)](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html) from version 2.1 of the Web Content Accessibility Guidelines for more information about why this is important and beneficial to fix.

The text of these links should all be unique. Adding the namespace's name would achieve that goal while providing additional clarity.

The new link text could be something like:
```gohtml
Limit results to the {{ $.Namespace }} namespace
```

All versions since v3.1.0 are affected by default. There are no steps necessary to reproduce beyond just "use Goldilocks."

### Version

v3.1.0

### Search

### Code of Conduct

### Additional context

_No response_

thewildmage:
This PR fixes #505.

## Checklist
* [x] I have signed the CLA

See #505.

Made the changes suggested in #505.

Different wording, maybe? This change is pretty straightforward.


sudermanjr (approved):
Where's the easy button? LGTM

PR #530, opened 2022-10-02
Make summary elements more accessible

### LINKED ISSUE #504: Summary elements shouldn't contain headings
### What happened?

Back when I did the big UI refresh, I foolishly put a bunch of `<h2>` elements inside `<summary>` elements. In some setups, this could result in the headings losing semantic meaning, an accessibility loss (see [MDN Docs: `<summary>`: The Disclosure Summary element: Summaries as headings](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/summary#summaries_as_headings)).

`<h2>` elements should be moved outside of the `<details>` elements, and new text should be provided for the `<summary>` elements.

This update also provides the opportunity to make the `<details>` elements simpler and more visually consistent in all places they're used by removing the custom styling and icons.

These changes might look something like this:
Mock-up of suggested changes.

All versions since v3.1.0 are affected by default. There are no steps necessary to reproduce beyond just "use Goldilocks."

### Version

v3.1.0+

### Search

### Code of Conduct

### Additional context

_No response_

thewildmage:
This PR fixes #504.

## Checklist
* [x] I have signed the CLA

To make the usage of `<summary>` HTML element more accessible.

There were two main impediments to accessibility in the current implementation:

1. There are `<summary>` elements that contain headings (i.e. `<h1>` through `<h6>`). This strips the headings of semantic meaning in some user setups. (This is the crux of Issue #504.)
2. The `<summary>` elements have custom open/close arrows, which alter semantic meaning and usability in some user setups.

This PR fixes both of these by moving the headings outside of the `<summary>` elements and going back to using the default open/close arrows. Notably, the default open/close arrows were already in use for the "Recommended Settings" sections (under Container info), so this change also makes things more consistent overall.

These changes are largely informed by "[The details and summary elements, again](https://www.scottohara.me/blog/2022/09/12/details-summary.html)" by Scott O'Hara, an accessibility specialist. The other suggestion there is to create fully custom open/close elements and not use `<details>` and `<summary>`. This alternative should be considered as it could make things even more usable and accessible than what this PR achieves. It'd be a bigger change though, which is why I've opted here for a more incremental improvement to start with.


pkg/dashboard/templates/container.gohtml
```diff
@@ -95,11 +95,9 @@
       </tr>
     </tbody>
   </table>
-
+
```
sudermanjr, line 98: nit: trailing whitespace

sudermanjr (approved):
awesome sauce. thanks!

pkg/dashboard/templates/container.gohtml
thewildmage: Oops. Fixed.

PR #531, opened 2022-10-02
Make headings only contain heading content

thewildmage:
This PR does not address any Issues.

## Checklist
* [x] I have signed the CLA

There are a few issues with the Guaranteed and Burstable QoS headings:

1. They contain cost data which isn't actually part of the heading. This is an accessibility issue because it's read as if it were part of the heading and shows up in the document outline when using a screen reader.
2. They contain `<abbr>` elements with `title` attributes that cause the headings to be read out by screen readers differently than they appear.
3. They link to the glossary. The more common convention for a heading containing a link is that it links to itself, enabling someone to jump back to that same spot on the page. Going against common convention can be confusing.

1. Moved the cost data out of the headings.
2. Removed the `<abbr>` elements. There's already a glossary on the same page.
3. Removed the glossary links.

These changes are meant to improve accessibility and make the HTML less complex. It's possible there are other ways to accomplish these goals, but I'm not sure what an alternative might be.

### DIFF
--- a/pkg/dashboard/assets/css/main.css
+++ b/pkg/dashboard/assets/css/main.css
@@ -402,6 +402,18 @@ COMPONENTS
   color: var(--color-accent-2);
 }
 
+.layoutCluster {
+  align-items: center;
+  gap: var(--gap--2);
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+}
+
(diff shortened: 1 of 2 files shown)


sudermanjr (approved):
Woo!

PR #544, opened 2022-10-22
Improve navigation accessibility and usability

thewildmage:
This PR does not address any specific open Issue.

## Checklist

* [x] I have signed the CLA

To improve the usability and accessibility of the inter- and intra-page navigation.

- Made the sidebar more consistent across all pages
- Hid the "Limit results to only X namespace" link when X is the only namespace shown
- Moved the filter out of the sidebar to a position adjacent to the content it affects
  - This should make it more clear what the filter does. It also makes it easier for someone navigating via keyboard to jump back and forth between the filter and the results
- Updated the filter form's styling to improve accessibility and better suit its new position
- Improved the accessibility of the filtering functionality with regard to screen readers
- Added accessible filter status updates (e.g. "Showing 1 out of 6 namespaces")
- Made a non-empty filter work on page refresh
- Handled the case where JavaScript isn't available (by accessibly hiding the filter form and leaving a static namespace count instead)
- Extracted JavaScript (JS) functions for testing for the availability of JS, showing elements, and hiding elements out into utilities that can be used across other JS functionality for improved consistency

These changes were largely informed by:

- "[Maybe Ignore type=search](https://adrianroselli.com/2019/07/ignore-typesearch.html)" by Adrian Roselli  
- "[Are we live?](https://www.scottohara.me/blog/2022/02/05/are-we-live.html)" by Scott O'Hara  
- "[The Many Lives of a Notification](https://www.youtube.com/watch?v=W5YAaLLBKhQ)" talk by Sarah Higley for A11yTalks (July 2020), and the [accompanying interactive slides](https://smhigley.github.io/slides/notification-lives), which are themselves publicly available code in [Sarah's `slides` repo](https://github.com/smhigley/slides/blob/master/notification-lives)  
- MDN's [ARIA live regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) reference

I believe based on this research that I've provided a solid solution, but there may be other possible approaches.

I've also tested these changes on MacOS with Firefox, Safari, and Brave (which is Chromium-based) using VoiceOver.


transient1:
Thanks for the PR, @thewildmage ! Excited to check out these changes.

sudermanjr (approved):
LGTM. Thanks!

PR #546, opened 2022-10-27
Fixed duplicated requests

hieptl:
This PR fixes #

## Checklist
* [x] I have signed the CLA

* Front-end is sending duplicated requests to oss-users API. We need to fix that issue.

* Removed the self-indentification.js file and its usage.

* None


sudermanjr (approved):
LGTM

PR #641, opened 2023-08-29
Show HPA status for workloads

### LINKED ISSUE #449: Show HPA Presence on Dashboard
**Is your feature request related to a problem? Please describe.**
When you have an HPA scaling on CPU or Memory, it can have unexpected effects on the Goldilocks recommendations. For example, if your HPA is set to 50% target CPU, then Goldilocks will always show a recommendation of about 50% of the request for a guaranteed QoS setting. In order to surface this interaction, it would be good to add some kind of indication to the dashboard that HPA is involved with this particular set of pods.

**Describe the solution you'd like**
- Add to the summary struct data about HPAs. Allow Goldilocks to collect that info when looking up current workload information
- Display this on the dashboard

**Describe alternatives you've considered**
There are a lot of ways this could be surfaced, but I can't think of any alternatives for the data collection part.

**Additional context**
See the containers from the couch episode on Goldilocks for a real world example.

sudermanjr:
Fixes #449

## Checklist
* [x] I have signed the CLA
* [ ] I have updated/added any relevant documentation

Since HPAs that scale on memory or CPU can affect the recommendations made by VPA, we want to surface that information to the end-user

Add an HPA to the summary, as well as to the cached objects in the cluster. If one is found, show info on dashboard. Also added an HPA section to the glossary

Lots of different ways to visualize. This seemed straightforward.


sudermanjr:
Screenshot 2023-08-29 at 3 53 24 PM

transient1:
My only feedback is re the screenshot. The "key" is a little hard to grok. It wasn't clear that `HPA: CPU Memory` meant that both were being used as metrics for scaling, nor that `(min 2/ max 5)` referred to the replicas in the HPA. I wonder what it would look like to take a little more real estate and give those columns/headings? As an aside, maybe it'd be cool to actually see what the HPA is set to here as well. 

This is an exciting feature! Looking forward to seeing how this progresses!

PR #667, opened 2023-11-03
Support Mi units

### LINKED ISSUE #522: Use mebibytes (Mi) instead of megabites (M) in memory recommendation
Goldilocks dashboard shows memory recommendation in megabytes units (M), which is 10^6 of bytes.
Mebibytes (Mi) instead use as base powers of 2.

Usually Kubernetes memory requests are specified in Mi and also different tools, like default Prometheus Operator Grafana dashboard show in MiB units.

As a feature option to switch between different types of measurements can be added.

From Kubernetes [docs](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/):
> Limits and requests for memory are measured in bytes. You can express memory as a plain integer or as a fixed-point integer using one of these suffixes: E, P, T, G, M, K. You can also use the power-of-two equivalents: Ei, Pi, Ti, Gi, Mi, Ki. For example, the following represent roughly the same value:
> 
> 128974848, 129e6, 129M, 123Mi

jetersen:
This PR fixes #522 

## Checklist
* [x] I have signed the CLA
* [ ] I have updated/added any relevant documentation

Add support for using BinarySI since due to the use of RoundUp converts the Memory units to DecimalSI

The main issue is RoundUp which converts the units to DecimalSI to do the rounding.
So I created a basic method that uses math to convert the unit back to BinarySI although due to the difference in the units there is a slight difference like 124M turns into 119Mi.
Perhaps a different approach is to simply replace K,M,G,T with Ki, Mi, Gi, Ti 🤔

See https://github.com/FairwindsOps/goldilocks/issues/522#issuecomment-1793018418

Perhaps there is a different approach to rounding 🤔
I tried other solutions but did not get the desired result from String method 😓

This applies to this PR:


cmd/dashboard.go
```diff
@@ -27,11 +27,12 @@ import (
...
-	showAllVPAs  bool
-	basePath     string
-	insightsHost string
-	enableCost   bool
+	serverPort        int
+	showAllVPAs       bool
+	basePath          string
+	insightsHost      string
+	enableCost        bool
+	useMemoryBinarySI bool
```
jetersen, line 35: Not sure about naming here 😅

cmd/dashboard.go
```diff
@@ -43,6 +44,7 @@ func init() {
 	dashboardCmd.PersistentFlags().StringVar(&basePath, "base-path", "/", "Path on which the dashboard is served.")
 	dashboardCmd.PersistentFlags().BoolVar(&enableCost, "enable-cost", true, "If set to false, the cost integration will be disabled on the dashboard.")
 	dashboardCmd.PersistentFlags().StringVar(&insightsHost, "insights-host", "https://insights.fairwinds.com", "Insights host for retrieving optional cost data.")
+	dashboardCmd.PersistentFlags().BoolVar(&useMemoryBinarySI, "binary-si", false, "Use binary SI units for memory (base 2) instead of decimal SI units (base 10).")
```
jetersen, line 47: Suggestion please, naming is hard 😓

pkg/utils/utils.go
```diff
@@ -77,21 +77,56 @@ func Difference(a, b []string) (diff []string) {
...
 		resource.Giga,
 		resource.Tera,
 	}
 	if mem, exists := rl[v1.ResourceMemory]; exists {
-		i := 0
-		maxAllowableStringLen := 5
-		for len(mem.String()) > maxAllowableStringLen && i < len(memoryScales)-1 {
+		maxLength := 5
+		if len(mem.String()) <= maxLength {
+			return rl
```
jetersen, line 120: Since RoundUp is the issue in this case I decided to simply return early here 🤔

pkg/utils/utils.go
```diff
@@ -77,21 +77,56 @@ func Difference(a, b []string) (diff []string) {
...
+	var scaleValue int64
+	switch scale {
+	case resource.Kilo:
+		scaleValue = Kibibyte
+	case resource.Mega:
+		scaleValue = Mebibyte
+	case resource.Giga:
+		scaleValue = Gibibyte
+	case resource.Tera:
+		scaleValue = Tebibyte
```
jetersen, line 97: Followed the scaled used below
```
resource.Kilo,
resource.Mega,
resource.Giga,
resource.Tera,
```

jetersen:
Not sure if it helpful to add things to the docs about the command flag for the dashboard? Perhaps in the helm chart?

jetersen:
The question becomes does this even need to be a command flag option. Since most tooling uses Mi and I wonder if we should simply replace the different units like `M` with `Mi` to avoid the Quantity math 🤔

sudermanjr:
Thanks for the PR! We will likely need some extra time to review this - it's an issue that has plagued us for a while :-D 

I do like making this a flagged feature to start, even if we plan to default differently later.

jetersen:
@sudermanjr @transient1 This definitely not stale 😓

Issue #686, opened 2024-01-12
Setting `--enable-cost=false` breaks the display of the yaml recommendations in the dashboard


apamildner:
### What happened?

When you set `--enable-cost=false` when calling the dashboard, the copy-paste feature of the requests
is no longer looking pretty (notice that the copy button is gone, and yaml is not pretty): 
<img width="564" alt="image" src="https://github.com/FairwindsOps/goldilocks/assets/10640074/fbcd8e8f-b400-4cde-8907-465ad506d44a">

### What did you expect to happen?

The yaml should display nicely like so:
<img width="494" alt="image" src="https://github.com/FairwindsOps/goldilocks/assets/10640074/cadf80c4-e9ba-4017-8ac7-a3cf451694d4">

### How can we reproduce this?

Set `--enable-cost=false` as a parameter to the dashboard

### Version

v4.10.0

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Issue #668, opened 2023-11-07
Wrong "current" shown for memory in dashboard


apamildner:
### What happened?

I'm trying to use goldilocks and see the recommended settings, this is the requests on my pod:
![image](https://github.com/FairwindsOps/goldilocks/assets/10640074/5915f363-0d31-42e6-8c71-c9781190d0bb)

This is how the VPA that was created looks:
![image](https://github.com/FairwindsOps/goldilocks/assets/10640074/0d0d44ff-a347-45fd-b90d-91e27b0f2192)

This is how the dashboard reports it:
![image](https://github.com/FairwindsOps/goldilocks/assets/10640074/adc8df12-79a1-4f61-869e-3eb28474e10f)

I don't think the memory value for "current" is correct, or what am I missing here?

### What did you expect to happen?

I expect the "current" settings to be the same as what I have put on the pod

### How can we reproduce this?

Hard to say, this is my values file for VPA:

```yaml

# This is important, its what makes it run only in recommender mode
updater:
  enabled: false

# This is important, its what makes it run only in recommender mode
admissionController:
  enabled: false

```

This is the values for goldilocks:
```yaml
vpa:
  enabled: false

metrics-server:
  # metrics-server.enabled -- If true, the metrics-server will be installed as a sub-chart
  enabled: false

controller:
  # controller.resources -- The resources block for the controller pods
  resources:
    limits:
      memory: 512Mi
    requests:
      cpu: 25m
      memory: 256Mi
dashboard:
  # dashboard.enabled -- If true, the dashboard component will be installed
  enabled: true

  # dashboard.excludeContainers -- Container names to exclude from displaying in the Goldilocks dashboard
  excludeContainers: "containerA,containerB"

  # dashboard.resources -- A resources block for the dashboard.
  resources:
    limits:
      memory: 512Mi
    requests:
      cpu: 25m
      memory: 256Mi

```

### Version

Versions: **Goldilocks helm chart**
```
repository       = "https://charts.fairwinds.com/stable"
chart = "goldilocks" 
version          = "8.0.0" 
```
**VPA** 
```
repository       = "https://charts.fairwinds.com/stable"
chart = "vpa"
version          = "3.0.2"
```

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

apamildner:
[...]

PR #687, opened 2024-01-12
Feature/make UI more customizeable and fix bugs

### LINKED ISSUE #673: Allow to disable CPU limit recommandation
**Is your feature request related to a problem? Please describe.**
According to a lot of documentation, it's not a good practive to defined CPU Limit as it can lead to CPU Throttling.
https://web.archive.org/web/20220805232857/https://home.robusta.dev/blog/stop-using-cpu-limits/

**Describe the solution you'd like**
It would be nice to have a way to disable recommandation of goldilcoks on it like an args on the dashboard that disable it.

**Describe alternatives you've considered**
None

**Additional context**
We would like to provide Goldlock to our dev. We have internally recommanded to not use in 99% of their case cpu limits.
If we give us the access to goldilock, he will actually made recommandation on CPU limits which is not what we want.

apamildner:
Thanks for a great tool for recommendations in kubernetes! I took some time to improve the code to suit our
organization better, and I hope you can consider merging at least some of these changes so we don't need to maintain our own fork and others can benefit from this too.

This PR fixes #673, #668, #686 

## Checklist
* [x] I have signed the CLA

## Description
This PR fixes some bugs that were present (see the issues linked). It also adds some UI customization options
to the dashboard to make it possible to limit what is shown for a give container. The changes should be fully backwards compatible.

The goal is to fix some bugs that were present that ruined a bit of the experience using this cool product! It also makes it easier to customize the UI to better fit the individual organizations using this.

* [x] Fix UI bug for yaml copy-paste recommendations even if `--enable-cost=false`
* [x] Fix UI bug where 1100Mi (or similar strings of len > 6) would be rendered in Ki instead of just 1100Mi due to a flaw in the roundup logic
* [x] Add option `--enable-cpu-limit-recommendation` which when false will no longer show CPU limit
* [x] Add option `--enable-burstable-recommendation` which when false will no longer show the "Burstable QoS"

The fix for the `1100Mi` (or similar strings of len > 6) is not optimal, my fix only pushes the problem forward (but writing for example 11000Mi is more unlikely to happen). Optimally we would round it _down_ to something like `1.1Gi` but I didn't have time to investigate how these conversions work in a good way.
There is also the issue that when we are not setting CPU limit, we are technically no longer in the QoS class guaranteed either. In general I think the QoS concept is more confusing than helpful in the dashboard and would see it removed, but I didn't want to make too big changes at this point :+1:

### DIFF
--- a/cmd/dashboard.go
+++ b/cmd/dashboard.go
@@ -31,8 +31,9 @@ var (
 	showAllVPAs  bool
 	basePath     string
 	insightsHost string
-	enableCost   bool
+	enableCost                    bool
 	enableCPULimitRecommendation  bool
+	enableBurstableRecommendation bool
 )
 
 func init() {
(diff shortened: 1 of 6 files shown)


bjuraga:
@apamildner any chance you can complete (sign) the CLA? It seems to be the only reason this feature was not merged, and it is needed by my team as well.

apamildner:
I think I maanged to sign it now. But I'm not using this tool any more, feel free to take my changes and incorporate in a branch of your own and attempt to get it merged. Since this was never merged here, I just maintained a fork until we stopped using it alltogether. I don't have capacity to get this branch updated now, my interests are focused elsewhere 👍

bjuraga:
thanks

Issue #534, opened 2022-10-06
Ability to ignore some controller by Kind


FrancoisPoinsot:
**Is your feature request related to a problem? Please describe.**
I am testing goldilocks.
I see it creates a VPA for each controller.
In my case I have a lot of Job and I would prefer to play it safe and not create VPA for those

**Describe the solution you'd like**
An envar/argument that would allow me to ignore some Controller by kind.
Such as `--ignore-controller-kind=Job,CronJob`

Maybe we would want some combinaison of namespace+kind

**Describe alternatives you've considered**
For now we can technically use RBAC to control what the goldilock controller can access
By removing the permission on specific Kind we can achieve something similar.
Goldilocks handles this error nicely.
But it does throw a lot of error logs.

rverchere:
Hi,

Is the ClusterRole already manage it?

When installing it using Helm, I have a ClusterRole with these rules, and job/cronjobs are not allowed:

```yaml
rules:
- apiGroups:
  - apps
  resources:
  - '*'
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - ""
  resources:
  - namespaces
  - pods
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - autoscaling.k8s.io
  resources:
  - verticalpodautoscalers
  verbs:
  - get
  - list
  - create
  - delete
  - update
- apiGroups:
  - argoproj.io
  resources:
  - rollouts
  verbs:
  - get
  - list
  - watch
```

Some logs:

```
goldilocks-controller-6f94c5d65f-62mg8 goldilocks E0301 16:06:21.022461       1 controller.go:229]  "msg"="Error retrieving parent object" "error"="cronjobs.batch is forbidden: User \"system:serviceaccount:vpa:goldilocks-controller\" cannot list resource \"cronjobs\" in API group \"batch\" in the namespace \"default\"" "v1beta1"="cronjobs"
goldilocks-controller-6f94c5d65f-62mg8 goldilocks E0301 16:06:21.022497       1 controller.go:147]  "msg"="An error occured retrieving the top level controller for this pod" "error"="cronjobs.batch is forbidden: User \"system:serviceaccount:vpa:goldilocks-controller\" cannot list resource \"cronjobs\" in API group \"batch\" in the namespace \"default\"" "my-job-1677685200-klwq5"="default"
[...]

PR #710, opened 2024-06-13
Ability to ignore some controller by Kind

k11h-de:
This PR fixes https://github.com/FairwindsOps/goldilocks/issues/534 

## Checklist
* [x] I have signed the CLA

adding a new arg `--ignore-controller-kind=Job,CronJob` that prevents certain controller kinds from being considered

## notes
I am sorry that I am not following your contribution guidelines. I am not a developer and was using GitHub Copilot to make these changes. The `make tidy` and `make build` are compiling without any errors - I have only limited resources and permissions on my works laptop. I would be grateful if you would still consider this improvement.

### DIFF
--- a/cmd/controller.go
+++ b/cmd/controller.go
@@ -28,6 +28,7 @@ import (
 
 var onByDefault bool
 var includeNamespaces []string
+var ignoreControllerKind []string
 var excludeNamespaces []string
 var dryRun bool
 
@@ -37,6 +38,7 @@ func init() {
 	controllerCmd.PersistentFlags().BoolVar(&dryRun, "dry-run", false, "If true, don't mutate resources, just list what would have been created.")
 	controllerCmd.PersistentFlags().StringSliceVar(&includeNamespaces, "include-namespaces", []string{}, "Comma delimited list of namespaces to include from recommendations.")
 	controllerCmd.PersistentFlags().StringSliceVar(&excludeNamespaces, "exclude-namespaces", []string{}, "Comma delimited list of namespaces to exclude from recommendations.")
+	controllerCmd.PersistentFlags().StringSliceVar(&ignoreControllerKind, "ignore-controller-kind", []string{}, "Comma delimited list of controller kinds to ignore from automatic VPA creation.")
 }
 
 var controllerCmd = &cobra.Command{
--- a/cmd/controller.go
+++ b/cmd/controller.go
@@ -38,7 +38,7 @@ func init() {
 	controllerCmd.PersistentFlags().BoolVar(&dryRun, "dry-run", false, "If true, don't mutate resources, just list what would have been created.")
 	controllerCmd.PersistentFlags().StringSliceVar(&includeNamespaces, "include-namespaces", []string{}, "Comma delimited list of namespaces to include from recommendations.")
 	controllerCmd.PersistentFlags().StringSliceVar(&excludeNamespaces, "exclude-namespaces", []string{}, "Comma delimited list of namespaces to exclude from recommendations.")
-	controllerCmd.PersistentFlags().StringSliceVar(&ignoreControllerKind, "ignore-controller-kind", []string{}, "Comma delimited list of controller kinds to ignore from automatic VPA creation.")
+	controllerCmd.PersistentFlags().StringSliceVar(&ignoreControllerKind, "ignore-controller-kind", []string{}, "Comma delimited list of controller kinds to exclude from recommendations.")
 }
 
 var controllerCmd = &cobra.Command{
--- a/cmd/controller.go
+++ b/cmd/controller.go
@@ -50,6 +50,7 @@ var controllerCmd = &cobra.Command{
 		vpaReconciler.OnByDefault = onByDefault
 		vpaReconciler.IncludeNamespaces = includeNamespaces
 		vpaReconciler.ExcludeNamespaces = excludeNamespaces
+		vpaReconciler.IgnoreControllerKind = ignoreControllerKind
 
 		klog.V(4).Infof("Starting controller with Reconciler: %+v", vpaReconciler)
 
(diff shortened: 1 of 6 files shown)


pkg/vpa/vpa.go
```diff
@@ -157,14 +158,19 @@ func (r Reconciler) namespaceIsManaged(namespace *corev1.Namespace) bool {
...
 	defaultMinReplicas, _ := vpaMinReplicasForResource(ns)
 
 	// these keys will eventually contain the leftover vpas that do not have a matching controller associated
 	vpaHasAssociatedController := map[string]bool{}
 	for _, controller := range controllers {
+		// Check if the controller kind is in the ignore list
+		if contains(ignoreControllerKind, controller.Kind) {
+			continue
+		}
+
```
sudermanjr, line 173: We would prefer to use https://github.com/samber/lo for this type of thing rather than writing one-off helper functions.

sudermanjr (changes requested):
Thanks for the PR.

Adding a new flag will require end-to-end tests for this flag at a minimum.

pkg/vpa/vpa.go
k11h-de: Thanks for your inout on how to improve the PR.
Could you please review the end-to-end test and change to use "github.com/samber/lo" instead the one-off helper function?

Thanks a lot

k11h-de: @sudermanjr 
Sorry, I am not so familiar with the work flows (and buttons) in Github - didn't meant to spam you.

pkg/vpa/vpa.go
```diff
@@ -51,6 +52,7 @@ type Reconciler struct {
 	DryRun                bool
 	IncludeNamespaces     []string
 	ExcludeNamespaces     []string
+	ignoreControllerKind  []string
```
sudermanjr, line 55: This field should be exported to match the other fields in this struct. This way it can be used by other packages

```suggestion
	IgnoreControllerKind  []string
```

cmd/controller.go
```diff
@@ -37,6 +38,7 @@ func init() {
 	controllerCmd.PersistentFlags().BoolVar(&dryRun, "dry-run", false, "If true, don't mutate resources, just list what would have been created.")
 	controllerCmd.PersistentFlags().StringSliceVar(&includeNamespaces, "include-namespaces", []string{}, "Comma delimited list of namespaces to include from recommendations.")
 	controllerCmd.PersistentFlags().StringSliceVar(&excludeNamespaces, "exclude-namespaces", []string{}, "Comma delimited list of namespaces to exclude from recommendations.")
+	controllerCmd.PersistentFlags().StringSliceVar(&ignoreControllerKind, "ignore-controller-kind", []string{}, "Comma delimited list of controller kinds to exclude from recommendations.")
```
sudermanjr, line 41: Below this, around line 52, this should be added to the vpaReconciler

pkg/vpa/vpa.go
```diff
@@ -110,7 +112,7 @@ func (r Reconciler) ReconcileNamespace(namespace *corev1.Namespace) error {
 		return err
 	}
 
-	return r.reconcileControllersAndVPAs(namespace, vpas, controllers)
+	return r.reconcileControllersAndVPAs(namespace, vpas, controllers, r.ignoreControllerKind)
```
sudermanjr, line 115: This field is already contained in the reconciler, no need to pass it like this.

```suggestion
	return r.reconcileControllersAndVPAs(namespace, vpas, controllers)
```

pkg/vpa/vpa.go
```diff
@@ -157,14 +159,19 @@ func (r Reconciler) namespaceIsManaged(namespace *corev1.Namespace) bool {
 	return r.OnByDefault
 }
 
-func (r Reconciler) reconcileControllersAndVPAs(ns *corev1.Namespace, vpas []vpav1.VerticalPodAutoscaler, controllers []Controller) error {
+func (r Reconciler) reconcileControllersAndVPAs(ns *corev1.Namespace, vpas []vpav1.VerticalPodAutoscaler, controllers []Controller, ignoreControllerKind []string) error {
```
sudermanjr, line 162: Same as above, should be in the Reconciler

```suggestion
func (r Reconciler) reconcileControllersAndVPAs(ns *corev1.Namespace, vpas []vpav1.VerticalPodAutoscaler, controllers []Controller) error {
```

pkg/vpa/vpa.go
```diff
@@ -157,14 +159,19 @@ func (r Reconciler) namespaceIsManaged(namespace *corev1.Namespace) bool {
...
 	defaultUpdateMode, _ := vpaUpdateModeForResource(ns)
 	defaultResourcePolicy, _ := vpaResourcePolicyForResource(ns)
 	defaultMinReplicas, _ := vpaMinReplicasForResource(ns)
 
 	// these keys will eventually contain the leftover vpas that do not have a matching controller associated
 	vpaHasAssociatedController := map[string]bool{}
 	for _, controller := range controllers {
+		// Check if the controller kind is in the ignore list
+		if lo.Contains(ignoreControllerKind, controller.Kind) {
```
sudermanjr, line 171:
```suggestion
		if lo.Contains(r.IgnoreControllerKind, controller.Kind) {
```

k11h-de:
Hi Andy,

Thanks a lot for the suggestions. I merged them.
Please let me know if you wish any other changes.

Sorry again for my poor coding skills.
Cheers and warm wishes from Berlin
Karsten

cmd/controller.go
sudermanjr: I don't believe this has been added to the actual instance of the struct that gets declared on lie 49

sudermanjr (changes requested):

cmd/controller.go
k11h-de: You're right. I made the requested change.

sudermanjr (approved):
Awesome. Thanks!

PR #713, opened 2024-06-21
Fixing BasePath for loading static assets and routing when a URL doesn't have a forward slash suffix

### LINKED ISSUE #652: basePath is not being honored for static assets, e.g. CSS and Images
### What happened?

setting the base path doesn't change the href for images and css

images and css to be downloaded and rendered

setg basePath

### Version

4.10.0

### Search

### Code of Conduct

### Additional context

_No response_

HereGoesName:
This PR fixes #652

## Checklist
* [x] I have signed the CLA

- Fixing basePath for loading static assets
- Fixing basePath for routing when the URL doesn't end in a `/`. This bug is mentioned in [this pull request comment](https://github.com/FairwindsOps/goldilocks/pull/471#issuecomment-1170408310)

Before this Pull Request static assets such as CSS and images don't load properly if the BasePath is set to anything other than the default. Additionally, the redirecting to the namespace URL doesn't happen if the BasePath is set and accessed via the URL without the forward slash suffix (e.g., a basePath of `example.com/foo` won't redirect to `example.com/foo/namespaces`, but `example.com/foo/` will).

- To fix basePath for loading static assets (without these changes the head's `<base href>` tag is ignored which causes the bug)
    - Modified html templates to refer to static resources without the `/` prefix
    - Added suffix of `/` to the URL path in validateBasePath() in dashboard.go
- In router.go I fixed the routing bug for basePaths without a forward slash at the end:
    - Moved functionality of router handling for the root URL in router.go to its own function rootHander()
    - Moved opts variable declaration to be a class-wide variable
    - Trimmed the `/` suffix from the PathPrefix when building the router in router.go
    - Added a router handler for "" to use the new rootHandler() method. This with the previous change enabled handling of root URLs without a forward slash
    - Removed from rootHandler() a check for if the URL.Path matches "/" as this is now redundant due to changes to opts.BasePath to fix the static assets bug
    - Added to rootHandler() a check for if the URL.Path matches the BasePath without its forward slash suffix. This prevents a URL like `example.com/foo` from returning a 404

It would be ideal not to use TrimSuffix() so often in this solution, however this may require changes to more files and code then I've done today

### Notes

Let me know if there's anything I've missed as I'm happy to make some adjustments

### DIFF
--- a/pkg/dashboard/router.go
+++ b/pkg/dashboard/router.go
@@ -50,8 +50,7 @@ func GetRouter(setters ...Option) *mux.Router {
 		setter(opts)
 	}
 
-	router := mux.NewRouter().PathPrefix(opts.BasePath).Subrouter()
-
+	router := mux.NewRouter().PathPrefix(opts.BasePath).Subrouter().StrictSlash(false)
 	// health
 	router.Handle("/health", Health("OK"))
 	router.Handle("/healthz", Healthz())
--- a/pkg/dashboard/router.go
+++ b/pkg/dashboard/router.go
@@ -17,6 +17,7 @@ package dashboard
 import (
 	"net/http"
 	"path"
+	"strings"
 
 	"k8s.io/klog/v2"
 
@@ -27,6 +28,8 @@ import (
 var (
 	markdownBox = (*packr.Box)(nil)
 )
+
+var opts *Options
 
 // GetMarkdownBox returns a binary-friendly set of markdown files with error details
 func GetMarkdownBox() *packr.Box {
@@ -45,12 +48,13 @@ func GetAssetBox() *packr.Box {
 
 // GetRouter returns a mux router serving all routes necessary for the dashboard
 func GetRouter(setters ...Option) *mux.Router {
-	opts := defaultOptions()
+	opts = defaultOptions()
 	for _, setter := range setters {
 		setter(opts)
 	}
 
-	router := mux.NewRouter().PathPrefix(opts.BasePath).Subrouter().StrictSlash(false)
+	router := mux.NewRouter().PathPrefix(strings.TrimSuffix(opts.BasePath, "/")).Subrouter()
+
 	// health
 	router.Handle("/health", Health("OK"))
 	router.Handle("/healthz", Healthz())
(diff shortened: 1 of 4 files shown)


pkg/dashboard/router.go
```diff
@@ -69,20 +72,23 @@ func GetRouter(setters ...Option) *mux.Router {
...
-			klog.Infof("404: %s", r.URL.Path)
-			http.NotFound(w, r)
-			return
-		}
-
-		klog.Infof("redirecting to %v", path.Join(opts.BasePath, "/namespaces"))
-		// default redirect on root path
-		http.Redirect(w, r, path.Join(opts.BasePath, "/namespaces"), http.StatusMovedPermanently)
-	})
+	router.HandleFunc("", rootHandler)
```
sudermanjr, line 75: Is this even necessary since we are adding a `/` to the basePath by default in the cmd/dashboard.go?

If it isn't, then I would prefer keeping the old flow without using a global variable.

sudermanjr, line 75: Is this actually necessary? I think I've always just used the "/" handler.

HereGoesName: Unfortunately if I don't add this new empty handler then from what I've seen a call to `example.com/foo` wouldn't redirect to the namespaces page, whilst a forward slash at the end would work. I added this so that the custom basePath would be inline with how the default works, as the default option redirects flawlessly without any issue with or without a forward slash at the end of the URL

sudermanjr: Weird. My gut says there's a better way to do all of this, but if it's working, then I'm good with it.

sudermanjr (approved):
Thanks!

pkg/dashboard/router.go
HereGoesName: I used mux's StrictSlash() in a previous commit but couldn't get it working. However, looking it up just now has made me realise that I used it wrong, and it seems that when using it the extra handler is unnecessary 😅
I'm implementing and testing this now, I should be able to get a commit out shortly to address this!

HereGoesName: I've simplified router.go in b849cd7 by reverting it mostly back to how it was

pkg/dashboard/router.go
```diff
@@ -75,7 +76,7 @@ func GetRouter(setters ...Option) *mux.Router {
 			klog.Infof("404: %s", r.URL.Path)
 			http.NotFound(w, r)
 			return
-		}
+			}
```
sudermanjr, line 79: Can you please run a `go fmt ./...` ?

sudermanjr (changes requested):

pkg/dashboard/router.go
HereGoesName: Whoops! Sorry about that, I missed that. I've addressed it in ccefc44

sudermanjr (approved):

PR #720, opened 2024-07-21
Added option to use VpaMinReplicasAnnotation from labels

devantler:
This PR adds the option of using `goldilocks.fairwinds.com/vpa-min-replicas: X` as a label.

## Checklist
* [x] I have signed the CLA

## Description

Currently it is not possible to set vpa-min-relplicas with labels, like it is with some of the other options. This fix should ensure that is also the case for `vpa-min-replicas`

To support labels for `vpa-min-replicas`

I added an if/else statement in the code where this was needed. The solution was inspired by how this is implemented other places in the code. Feel free to change it to your liking!

blank

### DIFF
--- a/pkg/vpa/vpa.go
+++ b/pkg/vpa/vpa.go
@@ -411,6 +411,8 @@ func vpaResourcePolicyForResource(obj runtime.Object) (*vpav1.PodResourcePolicy,
 	accessor, _ := meta.Accessor(obj)
 	if val, ok := accessor.GetAnnotations()[utils.VpaResourcePolicyAnnotation]; ok {
 		resourcePolicyStr = val
+	} else if val, ok := accessor.GetLabels()[utils.VpaResourcePolicyAnnotation]; ok {
+		resourcePolicyStr = val
 	}
 
 	if resourcePolicyStr == "" {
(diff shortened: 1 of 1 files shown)


pkg/vpa/vpa.go
```diff
@@ -436,6 +436,8 @@ func vpaMinReplicasForResource(obj runtime.Object) (*int32, bool) {
 	accessor, _ := meta.Accessor(obj)
 	if val, ok := accessor.GetAnnotations()[utils.VpaMinReplicasAnnotation]; ok {
 		minReplicasString = val
+	} else if val, ok := accessor.GetLabels()[utils.VpaMinReplicasAnnotation]; ok {
```
sudermanjr, line 439: Sorry, am I missing something here? This looks exactly the same as the previous if statement.

sudermanjr (changes requested):
I don't love only doing this for the minReplicas and not the other VPA settings. Can you make sure that all settings are updated?

devantler:
Sure, i'll have a look at it as soon as possible ☀️🕶️

devantler:
@sudermanjr I made a search for all cases that uses the `accessor.GetAnnotations()` method, and only found one more that was missing an `else` condition for retrieving the key/value from labels. The rest had both annotations and labels supported, where annotations is prioritized if both are set. 

I believe that should be all, so I have rerequested a review ✌🏻

sudermanjr:
Looks good. Now I need to figure out why the CI won't run

sudermanjr (approved):

PR #724, opened 2024-09-19
Allow the manual setting of recommenders on VPAs (not overwritten by the goldilocks controller) and a namespace label to add a custom recommender

brocoppler:
## Checklist
* [X] I have signed the CLA

To allow recommenders to be set on VPAs without being overwritten and provide a label that can be used to have goldilocks manage the setting of a given custom recommender

The spec is no longer always overwritten by the goldilocks controller, but rather it is additive when not explicitly a goldilocks managed field

1. The label can be the ONLY way to set a custom recommender if using goldilocks
2. No label can be provided by goldilocks, but allow recommenders to be added and not overwritten still

### DIFF
--- a/pkg/vpa/vpa.go
+++ b/pkg/vpa/vpa.go
@@ -22,9 +22,9 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/samber/lo"
 	"k8s.io/client-go/util/retry"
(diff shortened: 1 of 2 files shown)


brocoppler:
Accidentally committed with a different github account so going to close and reopen so the CLA only needs one signer

PR #725, opened 2024-09-20
Allow a custom recommender to be set via annotation or label on the namespace

brocoppler:
## Checklist
* [x] I have signed the CLA

## Description
What's the goal of this PR?
To allow recommenders to be set on VPAs without being overwritten and provide a label that can be used to have goldilocks manage the setting of a given custom recommender

What changes did you make?
The spec is no longer always overwritten by the goldilocks controller, but rather it is additive when not explicitly a goldilocks managed field

What alternative solution should we consider, if any?
The label can be the ONLY way to set a custom recommender if using goldilocks
No label can be provided by goldilocks, but allow recommenders to be added and not overwritten still

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -126,6 +126,11 @@ Example of an annotation
       "memory": "2048Mi" } }, { "containerName": "istio-proxy", "mode": "Off" }
       ] }
 ```
+#### VPA Custom Recommender
+
+If you want a specific namespace to use a custom VPA recommender,
+then you can label the workload with `goldilocks.fairwinds.com/vpa-recommenders=<recommender-name>`
+to add a recommender to the VPA specs for that namespace.
 
 #### Workload Specifications
 
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -129,7 +129,7 @@ Example of an annotation
 #### VPA Custom Recommender
 
 If you want a specific namespace to use a custom VPA recommender,
-then you can label the workload with `goldilocks.fairwinds.com/vpa-recommenders=<recommender-name>`
+then you can label or annotate the workload with `goldilocks.fairwinds.com/vpa-recommenders=<recommender-name>`
 to add a recommender to the VPA specs for that namespace.
 
 #### Workload Specifications
--- a/pkg/vpa/vpa.go
+++ b/pkg/vpa/vpa.go
@@ -221,7 +221,26 @@ func (r Reconciler) reconcileControllerAndVPA(ns *corev1.Namespace, controller C
 		klog.V(5).Infof("%s/%s has custom vpa-resource-policy", controller.Kind, controller.Name)
 	}
 
+	// Fetch the recommender name from the namespace label
+	recommenderName, err := vpaRecommenderNameForNamespace(ns)
+	if err != nil {
+		klog.Errorf("Error parsing recommender name for Namespace/%s: %v", ns.Name, err)
+		return err
+	}
+
 	desiredVPA := r.getVPAObject(vpa, ns, controller, vpaUpdateMode, vpaResourcePolicy, minReplicas)
+
+	// Set the recommender if recommenderName is not empty
+	if recommenderName != "" {
+		desiredVPA.Spec.Recommenders = []*vpav1.VerticalPodAutoscalerRecommenderSelector{
+			{
+				Name: recommenderName,
+			},
+		}
+	} else if vpa != nil {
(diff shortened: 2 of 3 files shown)


pkg/vpa/vpa.go
```diff
@@ -22,9 +22,9 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/samber/lo"
```
sudermanjr, line 25: this is unnecessary code churn, and no longer in alphabetical order, can we please revert this?

pkg/vpa/vpa.go
```diff
@@ -457,3 +484,11 @@ func vpaMinReplicasForResource(obj runtime.Object) (*int32, bool) {
 
 	return &minReplicasInt32, explicit
 }
+
+func vpaRecommenderNameForNamespace(ns *corev1.Namespace) (string, error) {
+	recommenderName, ok := ns.Labels["goldilocks.fairwinds.com/vpa-recommenders"]
```
sudermanjr, line 489: I believe we currently support both labels and annotations for the other settings that affect the VPA configuration. Would it be more appropriate to support both here?

pkg/vpa/vpa_test.go
```diff
@@ -16,6 +16,7 @@ package vpa
 
 import (
 	"context"
+	autoscaling "k8s.io/api/autoscaling/v1"
```
sudermanjr, line 19: Please follow the convention elsewhere in this repo and add this with the other k8s apis below, rather than in the built-in imports.

docs/advanced.md
```diff
@@ -126,6 +126,11 @@ Example of an annotation
       "memory": "2048Mi" } }, { "containerName": "istio-proxy", "mode": "Off" }
       ] }
 ```
+#### VPA Custom Recommender
```
sudermanjr, line 129: Thank you for adding docs, much appreciated

pkg/vpa/vpa.go
```diff
@@ -348,6 +367,7 @@ func (r Reconciler) getVPAObject(existingVPA *vpav1.VerticalPodAutoscaler, ns *c
 				Name:      "goldilocks-" + controller.Name,
 				Namespace: ns.Name,
 			},
+			Spec: vpav1.VerticalPodAutoscalerSpec{},
```
sudermanjr, line 370: is this added for a particular reason?

sudermanjr (changes requested):
I left some inline comments, but I also have some concerns about adding two completely separate pieces of functionality in a single PR.

- Setting the recommender makes complete sense, and I am happy to add this feature with some of the smaller issues fixed


- Not overwriting the current settings is something that needs further discussion. I know we have an open issue for "playing nicely with existing VPAs", however I believe that has been made obsolete by prefixing the VPA names with Goldilocks. Can you elaborate on the need for this, or link an issue that shows the desire/need for this change?

Ideally, these would be separate PRs.

pkg/vpa/vpa.go
brocoppler: I think that makes sense, I can update

pkg/vpa/vpa.go
brocoppler: Let me double check. I recall adding it when I added the label functionality, but it may have been a faulty test so Ill address in my second rev of this

brocoppler:
> * Setting the recommender makes complete sense, and I am happy to add this feature with some of the smaller issues fixed

I will update this PR to only include this

> * Not overwriting the current settings is something that needs further discussion. I know we have an open issue for "playing nicely with existing VPAs", however I believe that has been made obsolete by prefixing the VPA names with Goldilocks. Can you elaborate on the need for this, or link an issue that shows the desire/need for this change?

The idea was that if someone wanted to use goldilocks to get started and automatically have VPAs for a number of resources in a number namespaces, they could then fine tune some of those things manually (with recommeders being the inspiration for this) without goldilocks overwriting them or the need for new VPAs, but I understand there are pros and cons here and I think for my use case, the annotation/label is enough

brocoppler:
@sudermanjr I updated the PR to only include the new label/annotation and addressed all the minor things as well

brocoppler:
@transient1 @sudermanjr Hey just wanted to bump this again if you need anything else to get this merged

kfirsch:
I also need this feature. Any progress on it?

ledouxpl:
@brocoppler , @sudermanjr ... Any reason this was given-up upon? I'd be interested in this feature as well.

PR #783, opened 2025-08-15
Handle ETCd timeouts in large clusters

MonaFraise:
## Checklist
* [x] I have signed the CLA

We've been seeing timeouts and errors in Goldilocks due to the way AWS EKS control plane handles spiking and autoscaling. When we have lot of VPA objects, the EKS control plane cannot handle the constant requests from the Goldilocks controller.  Logs for this show up as `retrying of unary invoker failed` trying to connect to ETCd from the Kubernetes API and in Goldilocks as `Failed to store checkpoints. Reason: context deadline exceeded`. When Goldilocks is not able to store VPA changes, the next loop causes VPA to do all of the same work again.  In our experience, VPA starts kicking every changed deployment roughly every 30 seconds. This PR strives to improve the Goldilocks controller response to ETCd timeouts and control plane pressure. It watches for errors from the control plane, and backs off to let ETCd recover. There's also now a global rate limit to help keep the controller more consistent.

  1. Enhanced Controller Resilience:
    - Exponential backoff for watch stream failures and informer restarts
    - Control plane pressure detection with automatic pausing
    - Timeout handling for Kubernetes API calls
    - Classify retry able vs non retry able errors for logging

  2. Rate Limiting Improvements:
    - Implement intelligent backoff (1s, 2s, 4s, 8s, 15s, 30s max) for ETCd timeouts
    - Add backoffs for frequent informer restarts
    - Handle the DeletedFinalStateUnknown objects in event processing

  3. VPA Reconciliation Coordination:
    - Reconciliation loops complete fully to reduce control plane load

  4. Enhanced Error Handling:
    - Identify RBAC vs. control plane errors
    - Log timeouts with context and retry attempt counts
    - Add exponential backoff retry for API operations

  5. Timeout Management:
    - Add the timeout contexts! It fit with the rest of the project :)
    - Try to handle timeouts more consistently instead of spamming retries at a failing control plane

A global rate limit only - this would be simpler but less intelligent than the current approach which dynamically adjusts based on control plane health indicators and error patterns.

I set a reasonable default, but the rate limit is also configurable so we can scale Goldilocks properly in each cluster.

### DIFF
--- a/pkg/vpa/vpa.go
+++ b/pkg/vpa/vpa.go
@@ -630,8 +630,12 @@ func (r *Reconciler) IsReconciliationInProgress() bool {
 // WaitForReconciliationToComplete blocks until any active reconciliation completes
 func (r *Reconciler) WaitForReconciliationToComplete() {
 	// Acquire and immediately release a read lock to wait for active reconciliation
-	r.reconciliationMutex.RLock()
-	defer r.reconciliationMutex.RUnlock()
+	// Using an anonymous function to ensure immediate unlock without empty critical section
+	func() {
+		r.reconciliationMutex.RLock()
+		defer r.reconciliationMutex.RUnlock()
+		// The lock acquisition itself is the synchronization point we need
+	}()
 }
 
 


MonaFraise:
Thanks for getting the e2e tests going! I was able to fix the deadlock that kept VPAs from being created properly, and ensured that VPA objects are getting created correctly in all 4 k8s versions. Could you please kick off the e2e tests again?

PR #876, opened 2026-08-10
Do not panic on a VPA that has no targetRef

arpitjain099:
`VerticalPodAutoscalerSpec.TargetRef` is a pointer, and the VPA CRD does not mark it required, so the API server will happily accept a VPA with no `spec.targetRef`. Goldilocks dereferences it in two places without checking:

- `pkg/summary/summary.go:328` in `vpaMatchesWorkload`
- `pkg/summary/summary.go:155` when building the `workloadSummary`

`updateWorkloads` hits the first one, so any `GetSummary` call crashes as soon as one such VPA carries the goldilocks labels. The dashboard handler and the CLI summary both go through `GetSummary`.

Stack from a unit test that creates a labelled VPA with no `targetRef`:

```
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0]
summary.vpaMatchesWorkload(...)
	pkg/summary/summary.go:328
summary.(*Summarizer).updateWorkloads(...)
	pkg/summary/summary.go:315
summary.(*Summarizer).Update(...)
	pkg/summary/summary.go:259
summary.Summarizer.GetSummary(...)
	pkg/summary/summary.go:127
```

A VPA with no target does not point at any workload, so this treats it as matching nothing and logs and skips it in the summary loop, in the same style as the existing "no matching Workloads found for VPA/%s" path just below.

`Test_Summarizer_VPAWithoutTargetRef` is new: it panics on master and passes with the change. `go test ./pkg/...` is green and `gofmt` is clean.

### DIFF
--- a/pkg/summary/constants_test.go
+++ b/pkg/summary/constants_test.go
@@ -392,8 +392,7 @@ var testSummaryDaemonSet = Summary{
 	},
 }
 
-// A VPA whose spec omits targetRef, which the VPA CRD allows
-
+// testVPANoTargetRef is a VPA whose spec omits targetRef, which the VPA CRD allows
 var testVPANoTargetRef = &vpav1.VerticalPodAutoscaler{
 	ObjectMeta: metav1.ObjectMeta{
 		Name:      "test-vpa-no-target-ref",
--- a/pkg/summary/constants_test.go
+++ b/pkg/summary/constants_test.go
@@ -391,3 +391,18 @@ var testSummaryDaemonSet = Summary{
 		},
 	},
 }
+
+// A VPA whose spec omits targetRef, which the VPA CRD allows
+
+var testVPANoTargetRef = &vpav1.VerticalPodAutoscaler{
+	ObjectMeta: metav1.ObjectMeta{
+		Name:      "test-vpa-no-target-ref",
+		Labels:    utils.VPALabels,
+		Namespace: "testing",
+	},
+	Spec: vpav1.VerticalPodAutoscalerSpec{
+		UpdatePolicy: &vpav1.PodUpdatePolicy{
+			UpdateMode: &updateMode,
+		},
+	},
+}
--- a/pkg/summary/summary.go
+++ b/pkg/summary/summary.go
@@ -138,6 +138,11 @@ func (s Summarizer) GetSummary() (Summary, error) {
 	for _, vpa := range s.vpas {
 		klog.V(8).Infof("Analyzing vpa: %v", vpa.Name)
 
+		if vpa.Spec.TargetRef == nil {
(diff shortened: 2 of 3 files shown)


pkg/summary/summary.go
```diff
@@ -138,6 +138,11 @@ func (s Summarizer) GetSummary() (Summary, error) {
 	for _, vpa := range s.vpas {
 		klog.V(8).Infof("Analyzing vpa: %v", vpa.Name)
 
+		if vpa.Spec.TargetRef == nil {
+			klog.Errorf("no targetRef on VPA/%s", vpa.Name)
```
sudermanjr: nit: This feels more like a warning to me

sudermanjr:
Thanks for the great PR, and the tests. I'll get the pipeline going and hopefully get this merged.

sudermanjr (dismissed):

pkg/summary/constants_test.go
```diff
@@ -391,3 +391,18 @@ var testSummaryDaemonSet = Summary{
 		},
 	},
 }
+
+// A VPA whose spec omits targetRef, which the VPA CRD allows
+
+var testVPANoTargetRef = &vpav1.VerticalPodAutoscaler{
```
sudermanjr:
```suggestion
//testVPANoTargetRef is a VPA whose spec omits targetRef, which the VPA CRD allows
var testVPANoTargetRef = &vpav1.VerticalPodAutoscaler{
```

sudermanjr (changes requested):

pkg/summary/summary.go
arpitjain099: Agreed, changed to `klog.Warningf`. A VPA with no targetRef is a config mistake rather than something goldilocks got wrong, and it is skipped rather than failing the run, so a warning fits better.

pkg/summary/constants_test.go
arpitjain099: Taken, with one change: I had to keep the space after the slashes. Applying the suggestion exactly makes `gofmt` unhappy, since it normalizes doc comments:

```diff
-//testVPANoTargetRef is a VPA whose spec omits targetRef, which the VPA CRD allows
+// testVPANoTargetRef is a VPA whose spec omits targetRef, which the VPA CRD allows
```

So the comment now reads `// testVPANoTargetRef is a VPA whose spec omits targetRef, which the VPA CRD allows` on the line directly above the var, with the blank line dropped as you had it. `gofmt -l ./pkg/summary/` is clean and `go test ./pkg/...` still passes.

sudermanjr (approved):

pkg/summary/constants_test.go
sudermanjr: Ah yeah, my bad. Thanks!

